PR #15, opened 2024-01-12
Add ADR folder structure and templates

samdbmg:
# Details
Extends and supersedes #14 by adding documentation on how to run the ADR process. Also amends the PR template to make more sense in this repo.

# Pivotal Story
Story URL: https://www.pivotaltracker.com/story/show/186644493

# Related PRs
Supersedes #14

# Links to external test runs/working deployment
N/A

# Submitter PR Checks
_(tick as appropriate)_

- [X] Added bbc/rd-apmm-cloudfit team as a reviewer
- [ ] Tests exercise code appropriately
- [ ] New features and API breaks are flagged in commit messages using magic strings
- [ ] Repo maintainer is notified that a release is required
- [X] Documentation updated (README, Confluence, Docstrings, API spec, Engineering Guide, etc.)
- [ ] Downstream repos have been checked for potential breaks & fixed as needed
- [ ] APIs/UIs/CLIs updated as required
- [X] PR added to Pivotal story
- [ ] Any temporary code/configuration removed (e.g. test deployment environment, temporary commontooling branch)
- [ ] Any pins against pre-releases have been removed

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Tests exercise code appropriately
- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on Cloudfit PRs
- The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.
- For more details on how to asses PRs, see: https://github.com/bbc/rd-apmm-docs-ways-of-working/blob/master/workflow/PRChecklist.md

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,41 +1,31 @@
 # Details
 _Brief description of what this PR does._
 
-# Pivotal Story
+# Pivotal Story (if relevant)
 Story URL:
 
 # Related PRs
 _Where appropriate. Indicate order to be merged._
-
-# Links to external test runs/working deployment
-_Where appropriate, if separate to default CI run_
 
 # Submitter PR Checks
 _(tick as appropriate)_
 
 - [ ] Added bbc/rd-apmm-cloudfit team as a reviewer
 - [ ] PR completes task/fixes bug
-- [ ] Tests exercise code appropriately
-- [ ] New features and API breaks are flagged in commit messages using magic strings
-- [ ] Repo maintainer is notified that a release is required
-- [ ] Documentation updated (README, Confluence, Docstrings, API spec, Engineering Guide, etc.)
-- [ ] Downstream repos have been checked for potential breaks & fixed as needed
-- [ ] APIs/UIs/CLIs updated as required
-- [ ] PR added to Pivotal story
+- [ ] API version has been incremented if necessary
+- [ ] ADR status has been updated
+- [ ] Documentation updated (README, etc.)
+- [ ] PR added to Pivotal story (if relevant)
 - [ ] Follow-up stories added to Pivotal
-- [ ] Any temporary code/configuration removed (e.g. test deployment environment, temporary commontooling branch)
-- [ ] Any pins against pre-releases have been removed
 
 # Reviewer PR Checks
 _(tick as appropriate)_
 
 - [ ] PR completes task/fixes bug
-- [ ] Tests exercise code appropriately
 - [ ] Design makes sense, and fits with our current code base
 - [ ] Code is easy to follow
 - [ ] PR size is sensible
 - [ ] Commit history is sensible and tidy
 
-# Info on Cloudfit PRs
-- The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.
-- For more details on how to asses PRs, see: https://github.com/bbc/rd-apmm-docs-ways-of-working/blob/master/workflow/PRChecklist.md
+# Info on PRs
+The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,7 @@ _(tick as appropriate)_
 - [ ] Added bbc/rd-apmm-cloudfit team as a reviewer
 - [ ] PR completes task/fixes bug
 - [ ] API version has been incremented if necessary
-- [ ] ADR status has been updated
+- [ ] ADR status has been updated, and ADR implementation has been recorded
 - [ ] Documentation updated (README, etc.)
 - [ ] PR added to Pivotal story (if relevant)
 - [ ] Follow-up stories added to Pivotal
--- a/docs/adr/decisions/README.md
+++ b/docs/adr/decisions/README.md
@@ -1,0 +1,7 @@
+# Decisions
+
+This directory contains decision records for the tams api project.
+
+For new ADRs, please use [adr-template.md](adr-template.md) as basis - this is a variant of the MADR template.
+More information on MADR is available at <https://adr.github.io/madr/>.
+General information about architectural decision records is available at <https://adr.github.io/>.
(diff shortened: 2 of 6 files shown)


docs/adr/template/adr-template.md
```diff
@@ -0,0 +1,75 @@
+---
+# These are optional elements. Feel free to remove any of them.
+status: "{proposed | rejected | accepted | deprecated | … | superseded by [ADR-0005](0005-example.md)}"
```
samdbmg, line 3: I wonder how useful some of these statuses are? Would a better set be something like:
```
status: "{proposed | implemented | superseded by [ADR-0005](0005-example.md)}"
```

Where "proposed" means "we've decided to do this, but not built it yet", and "implemented" means "the current version of the API includes this change"?

robwadge: Indeed - "accepted | rejected" are catered for and documented by PR review audit trail.  May be useful to keep "deprecated" as an option?  Although it's likely that deprecation will involve superseding with a new ADR to explain the decision-making process.

AndrewGibb: "rejected" seems an important status for public domain discussion. We may find certain design decisions are questioned repeatedly, and having a "rejected" ADR to point at is better than producing a justification every time.

I'm not sure "implemented" is needed, as an "accepted" ADR can be updated with implementation details.

Your definition of "proposed" here implies that all the discussion takes place elsewhere, and only an accepted proposal makes it to an ADR. Line 11 in `Contributing.md` indicates that an ADR might be the start of a process which results in a change being included, or not. These two changes are not compatible.

j616: Alternative view: Is it worth keeping rejected. It can be as important to record the reasoning for not doing something as for doing something.

.github/pull_request_template.md
```diff
@@ -1,41 +1,31 @@
...
 - [ ] PR completes task/fixes bug
-- [ ] Tests exercise code appropriately
-- [ ] New features and API breaks are flagged in commit messages using magic strings
-- [ ] Repo maintainer is notified that a release is required
-- [ ] Documentation updated (README, Confluence, Docstrings, API spec, Engineering Guide, etc.)
-- [ ] Downstream repos have been checked for potential breaks & fixed as needed
-- [ ] APIs/UIs/CLIs updated as required
-- [ ] PR added to Pivotal story
+- [ ] API version has been incremented if necessary
+- [ ] ADR status has been updated
```
AndrewGibb, line 16:
```suggestion
- [ ] ADR status has been updated, and ADR implementation has been recorded
```

docs/adr/decisions/README.md
```diff
@@ -0,0 +1,33 @@
...
+1. Copy the [adr-template.md](../template/adr-template.md) into this directory
+2. Name it with the next number in sequence, and a sensible title (e.g. `00023-improve-source-hierarchy.md`)
+3. Fill in the template: at least the "Context", "Considered Options" and "Pros and Cons..." sections. Feel free to choose a preferred option and add it to the "Decision Outcome" section as well.
+4. If making the proposed changes (e.g. to the API, schemas or examples) helps to explain your proposal, make those changes as well
+5. Open a PR with your MADR document and any other changes
+
+## Reviewing the proposal
+During the review process, use PR comments to ask questions, make suggestions or generally discuss the proposal. Alternatively, if a synchronous discussion takes place (e.g. an in-person or online meeting), add a description of what was discussed and the outcome into the ADR document itself.
+
+Generally PRs should remain open for review for a period of time to invite comments from the wider TAMS community, and if necessary PRs should be circulated to interested individuals.
```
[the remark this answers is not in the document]
samdbmg: I'm trying not to bind us to something until we need to, on the basis that if it's urgent we'll start chasing. How about adding words to the effect that "The maintainers at BBC R&D will facilitate this process, and expedite it if necessary"

docs/adr/template/adr-template.md
samdbmg: **On "rejected"**
I'm inclined to agree with @AndrewGibb and @j616 here, where the subtlety is that a "rejected" ADR only gets merged if it's useful to retain the record. To use an example in the limit, a vexatious PR to make a stylistic change would just get closed, a seriously considered change to the model that we decide against would become "rejected"

**On "proposed"/"accepted"** 
> Your definition of "proposed" here implies that all the discussion takes place elsewhere, and only an accepted proposal makes it to an ADR. Line 11 in Contributing.md indicates that an ADR might be the start of a process which results in a change being included, or not. These two changes are not compatible.

I think the distinction is that once you've completed that process in Contributing.md, you have a PR for discussion, but the process of including or not including the change doesn't end until the document appears on `main`

Does that make sense?

**On "implemented"**
> I'm not sure "implemented" is needed, as an "accepted" ADR can be updated with implementation details.

I suppose the presence or absence of a PR link in the Implementation section serves that purpose, making it redundant.

AndrewGibb: > I think the distinction is that once you've completed that process in Contributing.md, you have a PR for discussion, but the process of including or not including the change doesn't end until the document appears on main

> Does that make sense?

That leads to the following set of conditions:

<table>
<tr>
 <td> PR status
 <td> adr status
 <td> actual meaning
<tr>
 <td> Not closed
 <td> "proposed"
 <td> A proposal, for discussion
<tr>
 <td> Closed
 <td> "proposed"
 <td> Proposal accepted
<tr>
 <td> Closed
 <td> "rejected"
 <td> Proposal rejected
</table>

Which doesn't make much sense to me. Including "accepted" for accepted proposals makes the meaning of the status much more clear.

docs/adr/decisions/README.md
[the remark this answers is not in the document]
AndrewGibb: Good point. Your suggestion describes the right kind of lightweight process too. Sounds good.

docs/adr/template/adr-template.md
samdbmg: Ok, I think the conclusion is we keep these as they are for now. Thanks for your insight everyone!

AndrewGibb (approved):
LGTM

PR #18, opened 2024-01-31
Add Sources to API

samdbmg:
# Details
- Adds an ADR document describing adding Sources as richer objects on the API
- Adds proposed implementation

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/186711160

# Related PRs
Followed by https://github.com/bbc/tams/pull/19

# Submitter PR Checks
_(tick as appropriate)_

- [x] Added bbc/rd-apmm-cloudfit team as a reviewer
- [X] PR completes task/fixes bug
- [X] API version has been incremented if necessary
- [x] ADR status has been updated, and ADR implementation has been recorded
- [X] Documentation updated (README, etc.)
- [X] PR added to Pivotal story (if relevant)
- [x] Follow-up stories added to Pivotal

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] PR completes task/fixes bug
- [ ] Design makes sense, and fits with our current code base
- [ ] Code is easy to follow
- [ ] PR size is sensible
- [ ] Commit history is sensible and tidy

# Info on PRs
The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.

PR #19, opened 2024-01-31
Indicate that timestamps are managed by the API server and ignored on PUT

samdbmg:
# Details
- Adds an ADR proposal describing that the `created` and various `updated` fields are ignored on PUT
- Adds corresponding wording to the field definitions

# Pivotal Story (if relevant)
Story URL: Came up while doing https://www.pivotaltracker.com/story/show/186711160

# Related PRs
Merge after https://github.com/bbc/tams/pull/18

# Submitter PR Checks
_(tick as appropriate)_

- [x] Added bbc/rd-apmm-cloudfit team as a reviewer

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/adr/decisions/0003-item-timestamps-managed-internally.md
+++ b/docs/adr/decisions/0003-item-timestamps-managed-internally.md
@@ -1,0 +1,50 @@
+---
+status: "proposed"
+---
+# Created and Modified Timestamps should be managed internally
+
+## Context and Problem Statement
+
+Flows (and Sources if [ADR 002](./0002-add-sources-to-api.md) is accepted) have fields indicating when they were created and last modified, however the API specification and schemas are not clear on whether these fields can be modified by a client, or whether they are exclusively managed by the server.
+
+See [schemas/flow-core.json](../../../api/schemas/flow-core.json) for definitions of those fields.
+They exist primarily to make it easier to find items of interest in the API and support engineering functions such as application development and debugging.
+
+## Considered Options
+
+* Option 1: Allow clients to change the dates when making a PUT request
+* Option 2: Ignore the date-time values in client requests
+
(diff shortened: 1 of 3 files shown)


docs/adr/decisions/0003-item-timestamps-managed-internally.md
```diff
@@ -0,0 +1,50 @@
...
+## Considered Options
+
+* Option 1: Allow clients to change the dates when making a PUT request
+* Option 2: Ignore the date-time values in client requests
+
+## Decision Outcome
+
+Chosen option: Option 2 - Ignore the date-time values in client requests, because it allows users to trust the behaviour of the date-time values, which is important given their primary purpose is to support development and debugging.
+
+This limits the use of the `created` field to refer to when the Flow was originally captured, but that information should ideally live elsewhere in a MAM system, or in the basic case, could be represented using tags.
```
philipnbbc, line 22: Not sure I understand this, but I assume with option 2 that the `created` field refers to when the Flow was stored rather than originally captured?

philipnbbc (approved):
Question inline and otherwise LGTM

docs/adr/decisions/0003-item-timestamps-managed-internally.md
samdbmg: Yes - as flow-core.json puts it:
> The date-time the flow was created in a given context, e.g. in a store.

With hindsight I don't like this very much - I can't see why you'd try and represent when the original media was originally captured using `created` anyway, so I've taken it out. If the time at which the first sample occurred is important, then you start the Flow timeline from there I think.

PR #35, opened 2024-03-22
Add flow collections to the API

robwadge:
# Details
ADR and associated implementation proposing addition of Flow collections to the API, to support multi-essence Flows

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187060669

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

- [x] PR completes task/fixes bug
- [x] PR added to Pivotal story (if relevant)
- [x] Follow-up stories added to Pivotal

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/examples/flow-get-200-multi.json
+++ b/api/examples/flow-get-200-multi.json
@@ -1,0 +1,24 @@
+{
+    "id": "e85efab4-993b-4ad6-9af3-4cd8d0d38860",
+    "source_id": "a77d0061-0878-4e8a-a114-772d03f952c1",
+    "format": "urn:x-nmos:format:multi",
+    "generation": 0,
+    "created": "2023-12-14T16:34:58.131620",
+    "tags": {
+        "ingested_by": "ingest_service_api"
+    },
+    "flow_collection": [
+        {
+            "id": "4f79cfd1-c057-47f4-8e4d-1b126ca7bf34",
+            "role": "video"
+        },
+        {
+            "id": "6101df05-06bb-41b8-8af4-cf7cd33df209",
+            "role": "audio"
+        },
+        {
+            "id": "e85efab4-993b-4ad6-9af3-4cd8d0d38860",
+            "role": "subtitles"
+        }
+    ]
+}
--- a/api/schemas/collection-item.json
+++ b/api/schemas/collection-item.json
@@ -1,0 +1,20 @@
+{
+    "type": "object",
+    "description": "Describes how an object (Source or Flow) is collected into another object of the same type",
+    "title": "Collection Item",
+    "required": [
+        "id",
+        "role"
+    ],
+    "properties": {
+        "id": {
+            "description": "Source or Flow Identifier of the member of this collection.  Must already be registered in TAMS",
+            "type": "string",
+            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
+        },
+        "role": {
+            "description": "A human-readable role of the element in this collection (e.g. 'R' to denote a right audio channel in a collection of mono audio Sources, or )",
+            "type": "string"
+        }
+    }
+}
--- a/api/schemas/source.json
+++ b/api/schemas/source.json
@@ -19,7 +19,7 @@
                 "urn:x-nmos:format:video",
                 "urn:x-nmos:format:audio",
                 "urn:x-nmos:format:data",
-                "urn:x-nmos:format:mux"
+                "urn:x-nmos:format:multi"
             ],
             "format": "uri"
         },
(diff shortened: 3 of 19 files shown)


api/schemas/collection-item.json
```diff
@@ -0,0 +1,25 @@
+{
+    "type": "object",
+    "description": "Describes how a Flow is collected into another (e.g. multi-essence) Flow",
```
samdbmg, line 3: Nit: Genericisation
```suggestion
    "description": "Describes how an object (Source or Flow) is collected into another (e.g. multi-essence) object of the same type",
```

docs/adr/decisions/0012-add-flow-collections.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Decision Outcome
+
+Chosen option: Option 3: Add multi-essence Flows that register collections of Flows in the TAMS API, because
+
+1. This facility is needed for even the most basic use cases
+2. The TAMS API needs to be self-contained for simple, minimal configurations
+3. Adding collections at the Source level doesn't provide all the information required to describe the corresponding relationships at the Flow level
+
+Extending this to add Option 3a would be desirable, but could be added later
```
samdbmg, line 33: Looking at the implementation, isn't it 3a anyway? Although it might be that practical stores don't implement the inferred relationships and have it as a "known limitation" initially

api/schemas/source.json
```diff
@@ -51,6 +51,21 @@
         },
         "tags": {
             "$ref": "tags.json"
+        },
+        "source_collection": {
+            "type": "array",
+            "description": "List of Sources that are collected together by this Source. This attribute is intended to be read-only. Implementations SHOULD ignore this if given in a PUT request, and instead manage it internally. Source collections can be inferred from Flow collection definitions.",
```
samdbmg, line 57: How do I construct a custom group in this model? Say I was delivered a package of video, English audio and English subtitles, and I then commissioned Welsh subtitles. Can I construct a Source that refers to video, English audio, Welsh subs without generating an asset containing all three (e.g. because I want to create the asset later)? Do I just create an empty Flow collecting the right Flows together, and then "make it real" at the time I actually need it?

samdbmg (commented):
Looks good overall - I've inlined a few nits and a question about creating new Sources

docs/adr/decisions/0012-add-flow-collections.md
robwadge: That was my intention - proposed implementation is 3a but implementers could choose to limit to option 3.  Could change the ADR to make this clearer.

api/schemas/source.json
robwadge: The application note (to follow) will cover this, but essentially yes, what you propose is more or less how you'd achieve this.

docs/adr/decisions/0012-add-flow-collections.md
```diff
@@ -0,0 +1,90 @@
+---
+status: "proposed"
+---
+# Add collections to flow and source metadata schemas
+
+## Context and Problem Statement
+
+For optimum flexibility, media Flows (and their parent Sources) are defined as mono-essence.
+However, for all practical purposes, a means of describing how separate mono-essence Flows relate together is needed.
```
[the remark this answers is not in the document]
robwadge: Yes that was intentional.  Thought it would make it easier to read, but clearly not!

api/examples/flow-get-200-multi.json
```diff
@@ -0,0 +1,29 @@
...
+    "id": "e85efab4-993b-4ad6-9af3-4cd8d0d38860",
+    "source_id": "a77d0061-0878-4e8a-a114-772d03f952c1",
+    "format": "urn:x-nmos:format:data",
+    "generation": 0,
+    "created": "2023-12-14T16:34:58.131620",
+    "tags": {
+        "ingested_by": "ingest_service_api"
+    },
+    "codec": "application/ttml+xml",
+    "container": "application/ttml+xml",
```
philipnbbc, line 11: `video/mp2t` would work better in this example. However, this is the container in the TAMS and as this is a multi-essence flow there may not be a container if there is no media object, i.e. mono-essence is stored. The `container` property would need to be optional as well.

api/schemas/collection-item.json
```diff
@@ -0,0 +1,25 @@
...
+    "title": "Collection Item",
+    "required": [
+        "id",
+        "slot",
+        "role"
+    ],
+    "properties": {
+        "id": {
+            "description": "Source or Flow Identifier of the member of this collection",
```
philipnbbc, line 12: Is the member required to be registered in the TAMS?

api/schemas/source.json
philipnbbc: I think it seems nicer to be able to populate a Source structure without resorting to a side-effect of creating empty / placeholder Flows. The API has an endpoint to PUT a Source and so it seems reasonable to allow the same for a multi-essence source? That still requires a multi-essence Flow to be created to specify which flows were grouped.

I suppose part of this relates to what we decide (have decided, don't remember) happens if a Flow is deleted; is the Source deleted as well if no more Flows reference it? I.e. do we consider the Sources to be a view only on Flows in the TAMS.

api/schemas/source.json
```diff
@@ -19,7 +19,7 @@
                 "urn:x-nmos:format:video",
                 "urn:x-nmos:format:audio",
                 "urn:x-nmos:format:data",
-                "urn:x-nmos:format:mux"
+                "urn:x-nmos:format:multi"
```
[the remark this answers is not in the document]
robwadge: We're in discussion with AMWA about this, but the situation is further complicated by the changes to IETF policy around URN namespace registration.  The best course of action is to leave it like this for now and come back to fix it up later.  Can confirm that `format:multi` is unique in the x-nmos namespace, at least.

philipnbbc (commented):
A couple of my comments inline

api/schemas/collection-item.json
robwadge: Yes.  I'll make that clear in the spec.

philipnbbc (approved):
LGTM. As discussed, there are some complications around grouping conflicts and what to do if Flows get deleted (Source inference conflicts, missing slots, etc.), but that's something we can figure that out as we implement these changes.

I was surprised the the JSON schema validator was ok with the `codec` and `container` properties being on flow-core and only `required` in flow-video etc., but it turns out that is [valid according to a JSON schema editor](https://github.com/OAI/OpenAPI-Specification/issues/1608#issuecomment-398607898).

PR #36, opened 2024-03-22
Clarify expectations of Flow Segment get_url

samdbmg:
# Details
Adds an ADR proposing some clarifications on how the `get_url` field on Flow Segments work:
- Makes `get_url` mandatory in Flow Segment responses
- Clarifies expectations around credentials when making segment requests
- Replaces `get_url` with a list of `get_urls`

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187054769

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ The media data contained within Flow Segments may be stored separately from the 
 For example our implementation uses a database (_e.g._ Amazon DynamoDB) to store Flow Segment metadata and an object store (_e.g._ AWS S3) to store the media data for Flow Segments.
 We refer to media data stored in the object store as 'media objects'.
 The Flow Segment has a list of S3 download urls which are the location of the media object(s) that contains the stored media data for the Flow Segment.
-If multiple URLs are returned in the list, they are assumed to be identical.
+If multiple URLs are returned in the list, they are assumed to return identical media data.
 When writing to the store, the S3 URLs can be passed to a client permitting them to upload media data directly.
 
 Another advantage of separating the media data and metadata planes in this way is that a particular Flow Segment can be referenced by multiple flows.
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ If multiple URLs are returned in the list, they are assumed to be identical.
 When writing to the store, the S3 URLs can be passed to a client permitting them to upload media data directly.
 
 Another advantage of separating the media data and metadata planes in this way is that a particular Flow Segment can be referenced by multiple flows.
-On the metadata side, the Flow Segment is just a URL, so any number of flows can record that same URL against other `<flow_id, timestamp>` tuples.
+On the metadata side, the Flow Segment is just an object ID, so any number of flows can record that same ID against other `<flow_id, timestamp>` tuples.
 This allows for copy-on-write semantics: immutability means a new Flow must be created to make changes to existing parts of the timeline, but for unmodified portions of the timeline the new `<flow_id, timestamp>` tuple points to the existing object ID or a part of it.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -54,11 +54,12 @@ The store provides a mechanism to upload and register new Flow Segments, and an
 The media data contained within Flow Segments may be stored separately from the metadata linking them to a position on the timeline, separating the media data and metadata planes.
 For example our implementation uses a database (_e.g._ Amazon DynamoDB) to store Flow Segment metadata and an object store (_e.g._ AWS S3) to store the media data for Flow Segments.
 We refer to media data stored in the object store as 'media objects'.
-The Flow Segment has a single S3 download url which is the location of the media object that contains the stored media data for the Flow Segment.
+The Flow Segment has a list of S3 download urls which are the location of the media object(s) that contains the stored media data for the Flow Segment.
+If multiple URLs are returned in the list, they are assumed to be identical.
```
philipnbbc, line 58: `they` refers to the media object?
```suggestion
If multiple URLs are returned in the list, they are assumed to return identical media data.
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -921,6 +925,14 @@ paths:
                     subset of the samples within using `ts_offset` for the second segment in the sequence.
                   value:
                     $ref: examples/flow-segments-get-200-sample-offset.json
+                multiple_urls:
+                  summary: Multiple URLs
+                  description: |
+                    This example shows how the media object for a Flow Segment could be accessed using multiple URLs.
+                    Clients are free to choose any of the URLs given (which are all identical), however the server has
```
philipnbbc, line 932: Same as before
```suggestion
                    Clients are free to choose any of the URLs given (which return identical media objects), however the server has
```

philipnbbc (approved):
Minor comprehension comment about what "identical" is referring to on a first read (happy for you to keep as-is). Otherwise LGTM

johnbilt:
The basic API structure looks good - nice to see labels to identify what the end points are.  However at the risk of opening a can of worms what happens about adding or removing URLs?

samdbmg:
> What happens about adding or removing URLs after the original segment record has been created?

On a "keep the core API simple" basis, I'd be inclined to say it should be an out-of-band extra that's neither required nor expressly forbidden. From a client's perspective, the list of `get_urls` they receive is a snapshot, only valid at that point in time, and may change if they make another request, but the semantics ("all URLs are equally valid and interchangeable") still apply.

From a server's perspective, it depends how you've implemented it. Either way you have to provide a `put_url` and an `object_id` to the client when they write content. The server might then choose to replicate that content somewhere else (another region?) and it can then generate `get_urls` covering both regions. Or it might be there's a CDN and/or an on-site cache involved, and it generates a pair of `get_urls`: one to go via the CDN, and one via the local cache. Or it might be you have A and B resilience ingests, leading to A and B Flows (of the same Source) which are defined to be bit-for-bit identical, and your particular TAMS store implementation responds to a signal that turns that into a separate resilient pair Flow C. Then the store lists URLs for both Flow A objects and Flow B objects in Flow C `get_urls` (unless one segment is missing, in which case you only get one: that's what the resilience was for!)

The other way to approach it would (potentially) be to allow registering additional duplicate Flow Segments (producing additional URLs), but the trade-off is it makes it easier to break immutability, because a client could come along and register a duplicate segment that differs from the original, and it would be difficult to enforce against that! IMO, we should leave it as is until the need arises to relax that restriction

PR #39, opened 2024-04-02
Add event stream

samdbmg:
# Details
Adds an ADR proposing a couple of possible approaches to an event notification stream, along with a proposal about what the message format might look like.

I think Option 3 (telemetry) probably doesn't make sense, but I'd appreciate thoughts on Option 1 (specify notification mechanism) vs Option 2 (leave it up to implementations). For now I've done the implementation for Option 2, but Option 1 is a build on that.

Also, it might seem like [AsyncAPI](https://www.asyncapi.com/) is the better fit for describing this, but it's quite high-friction for both HTTP webhooks and AWS EventBridge, which are the two possible implementations we've previously talked about, so using OpenAPI seemed the better approach for a spec.

I'd also appreciate thoughts on how Sources should work - see comments in the OpenAPI doc.

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187071067

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,14 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
 
 ![Graphic showing the Flow timeline and Flow Segments of Flows X, Y and Z, where Z is composed of a mix of re-used segments and new media](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ.drawio.png)
 
+### Events from the API
+
+The TAMS API specifies a list of event messages that should be generated and sent to interested clients in response to certain events, such as the creation of a new Flow.
+This is intended to reduce the amount of polling required by clients to keep up with activity on a store by providing a way to push updates about content they have access to instead.
+
+However the specification is deliberately left open-ended; only the message bodies are specified, but not the protocol by which they are carried nor the method by which clients subscribe.
+It is assumed that implementations will provide a suitable mechanism, such as a call to allow clients to subscribe to webhooks, or details of an event bus to connect to and receive the messages.
+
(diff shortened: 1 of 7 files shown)


philipnbbc (approved):
Some comments inline and otherwise LGTM.

I agree that Option 1 or 2 makes sense over Option 3. I think the structure of the message format will probably change for different delivery mechanisms and so I see it's more about getting maximum commonality between implementations. Similar to the HTTP object store choice, maybe a message format plus a implementation would help get some form consensus of what is needed? Something like Option 2 plus a simple example implementation, but not going as far as Option 1 just yet.

samdbmg:
> Similar to the HTTP object store choice, maybe a message format plus a implementation would help get some form consensus of what is needed? Something like Option 2 plus a simple example implementation, but not going as far as Option 1 just yet.

How's [8dcf44d](https://github.com/bbc/tams/pull/39/commits/8dcf44dd9f7db01d6b23e7ab0a4bacd7df22dcd5) for that? 
Downside is it adds a bit of complexity to the API and it's a bit difficult to specify adequately in schema without being unduly prescriptive. But I think making it optional mitigates that a bit? Plus webhooks come with quite a few additional security considerations - I found https://webhooks.fyi/best-practices/webhook-providers#implement-security-on-egress-communication to be interesting reading.

philipnbbc (approved):

PR #45, opened 2024-04-25
Add ADR for checksum and filesize metadata

j616:
# Details
Add ADR for checksum and filesize metadata

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187060666

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/adr/0016-checksums-and-filesize.md
+++ b/docs/adr/0016-checksums-and-filesize.md
@@ -1,5 +1,5 @@
 ---
-status: "proposed"
+status: "accepted"
 ---
 # Add Object Checksums and Filesizes
 
--- a/docs/adr/0016-checksums-and-filesize.md
+++ b/docs/adr/0016-checksums-and-filesize.md
@@ -1,0 +1,71 @@
+---
+status: "proposed"
+---
+# Add Object Checksums and Filesizes
+
+## Context and Problem Statement
+
+TAMS clients need to have confidence in the integrity of media, that it has been transferred without corruption, and that the file transferred is the one that is expected.
+
+Common methods of achieving this are `filesize` and `checksum` metadata.
+Filesize provides a very basic method of verifying the file.
+Is it the length that is expected?
+Has the full file been transferred?
+Checksums provide are slightly more expensive computationally, but provide a more reliable method of validating the file.
+Cryptographically, is the file highly likely to be identical to the expected file.
+
+This ADR considers if `filesize` and `checksum` metadata should be added to the TAMS API.
+
+## Decision Drivers
+
+* Provide high confidence in file integrity
+* Provide a detection method for unauthorised/malicious file tampering
+
+## Considered Options
+
+* Option 1: Provide metadata fields for `filesize` and `checksum` which are populated by the client on write
+* Option 2: Require TAMS implementations to calculate and populate these fields on write
+* Option 3: Rely on existing `filesize` and `checksum` functionality in storage layers such as AWS S3
+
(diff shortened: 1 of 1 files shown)


philipnbbc (commented):
Should TAMS expose these metadata through the API, even though it or the writer clients don't generate them? How does a client get this metadata? Would a `head_url` (e.g. a pre-signed HEAD request) in the Flow Segment work if the storage backend includes these metadata in the HTTP headers?

j616:
@philipnbbc For S3, [x-amz-checksum-mode](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax) should be set in the `GET` or `HEAD` request to the Object. I'm not a fan of adding this functionality to TAMS itself as it will increase the number of requests the API itself will have to make. One additional request per Object. The alternative is that this metadata is obtained by the client in the same request as the data. Additionally, supporting this in the API would only help with verifying downloads. As described in the ADR, verifying uploads would have to be accomplished before registering the segment. For S3, this would most likely by accomplished by [sending the checksum alongside the data](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestBody in the `PUT` request. Again, this would allow the verification and data transfer to happen in the one request. If we verified the data on `PUT` by checking the TAMS API after the fact, this could cause issues in the case that the upload is corrupted and there is a checksum miss-match. The object would have to be registered before you could retrieve the checksum and could be made available in it's corrupted state.

philipnbbc:
> @philipnbbc For S3, [x-amz-checksum-mode](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax) should be set in the `GET` or `HEAD` request to the Object. I'm not a fan of adding this functionality to TAMS itself as it will increase the number of requests the API itself will have to make. One additional request per Object. The alternative is that this metadata is obtained by the client in the same request as the data. Additionally, supporting this in the API would only help with verifying downloads. As described in the ADR, verifying uploads would have to be accomplished before registering the segment. For S3, this would most likely by accomplished by [sending the checksum alongside the data](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestBody in the `PUT` request. Again, this would allow the verification and data transfer to happen in the one request. If we verified the data on `PUT` by checking the TAMS API after the fact, this could cause issues in the case that the upload is corrupted and there is a checksum miss-match. The object would have to be registered before you could retrieve the checksum and could be made available in it's corrupted state.

OK.  If we find that a `HEAD` request is useful to retrieve checksum and size headers then we could extend the API to have either a pre-signed HEAD request labelled "get_url", a separate `head_urls` list or a specific `head_url`.

Similarly for the PUT request, we would need to allow the client to include checksum headers in the pre-signed PUT request. It's not 100% clear to me whether a presigned URL can be created because the `x-amz-checksum-sha256` is only known / calculated by the client.

philipnbbc:
> Similarly for the PUT request, we would need to allow the client to include checksum headers in the pre-signed PUT request. It's not 100% clear to me whether a presigned URL can be created because the `x-amz-checksum-sha256` is only known / calculated by the client.

philipnbbc (approved):

PR #56, opened 2024-05-16
Update documentation links to use new versioned documentation URLs

j616:
# Details
Update documentation links following change to endpoints introduced in https://github.com/bbc/tams/pull/55

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187576124

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,17 +70,17 @@ The Flow model is aligned with the principles and schemas of [AMWA NMOS IS-04](h
 The process of reading from the store is:
 
 1. Client identifies the Flow ID and timerange of interest
-2. Client makes a request to [`GET flows/<flow_id>/segments?timerange=<timerange>`](https://bbc.github.io/tams/#/operations/GET_flows-flowId-segments) and receives a list of Flow Segments, including their timeranges and download URLs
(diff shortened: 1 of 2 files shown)


philipnbbc (approved):
Would `main` be a better target than `4.0` or should we update these with each version change? I guess most people looking at the README.md on GitHub will be on the main branch and therefore the links in most cases will be correct if these were using main.

Otherwise LGTM

j616:
@philipnbbc My concern with `main` is that the documentation updated is "how-to" type documentation. I'd expect this would be used against an existing deployed instance which would likely be the latest version, rather than a pre-release. 

There's an argument that we should also have a `latest` endpoint for documentation which is the latest released version. That looked like it might take a little more faffing than needed for this first pass. But we could consider it in future. 

I guess arguably, this sort of documentation should also be packaged into the published bundle so how-to's are there for each release. But again, that felt a bit beyond this initial scope.

PR #57, opened 2024-05-30
Replace "Flow Deletion Request" in description with "Flow Delete Request"

j616:
# Details
Fixes a typo referring to a "Flow Deletion Request" which should say "Flow Delete Request"

# Pivotal Story (if relevant)
Story URL:

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -762,7 +762,8 @@ paths:
     delete:
       description: |
         Deletes the flow and associated segments. If flow segment deletion
-        takes too long then this request will redirect to a Flow Delete Request via a "Location" header
+        takes too long then this request will return 202 Accepted and the `Location` header will point to a
+        Flow Delete Request to monitor deletion progress
       operationId: DELETE_flows-flowId
       tags:
         - Flows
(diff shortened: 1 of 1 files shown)


api/TimeAddressableMediaStore.yaml
```diff
@@ -762,7 +762,7 @@ paths:
     delete:
       description: |
         Deletes the flow and associated segments. If flow segment deletion
-        takes too long then this request will redirect to a Flow Deletion Request
+        takes too long then this request will redirect to a Flow Delete Request via a "Location" header
```
samdbmg, line 765: The spec has this as returning a `202` with a `Location` header, but the implementation in Squirrel returns a 201 - I'm leaning towards Squirrel being incorrect here, and I've added a note to https://www.pivotaltracker.com/story/show/187691389 to fix that.

Nit: Not sure I like calling this a "redirect" since that implies a 3xx code. How's this?
```suggestion
        takes too long then this request will return 202 Accepted and the `Location` header will point to a
        Flow Delete Request to monitor deletion progress
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1151,7 +1151,7 @@ paths:
     delete:
       description: |
         Deletes the flow segments. If the deletion takes too long then this
-        request will redirect to a Flow Deletion Request.
+        request will redirect to a Flow Delete Request via a "Location" header.
```
samdbmg, line 1154: Nit: as above
```suggestion
        takes too long then this request will return 202 Accepted and the `Location` header will point to a
        Flow Delete Request to monitor deletion progress
```

samdbmg (approved):
LGTM - one nit inlined

PR #64, opened 2024-06-12
Audit of HTTP responses and schemas

j616:
# Details
This PR is the result of an audit of response codes and schema's. This PR:
* Removes 400 responses from queries which cannot return them
* Adds 400 responses to queries that can return them but were missing them
* Adds 404 responses for invalid path parameters
  * 404 is used here because invalid path parameters mean the path doesn't match the path pattern in the spec and, thus, the path is not found
* Adds minimum and maximum values to numeric parameters in the json schemas where sensible
  * Note that in all these cases, negative or zero values are nonsensical. These minimum/maximums reflect what would already be considered bugs in implementations.
* Adds notes in parameter descriptions indicating where one parameter is required based on the value of another
* Removes defaults in json schemas which cause unexpected behaviour in machine-generated implementations
* Expand some descriptions
* Adds missing schemas for flow tags endpoints
* Uses `additionalProperties` instead of `patternProperties` for tags schema to provide better rendered documentation, and better support machine-generated implementations

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187691389

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -368,7 +368,7 @@ paths:
         "200":
           $ref: '#/components/responses/trait_resource_info_head_200'
         "404":
-          $ref: '#/components/responses/trait_resource_info_head_404'
+          description: The requested Source or tag does not exist.
     get:
       summary: Source Tag Value
       description: Return the tag value associated with the tag name.
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -327,6 +327,8 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/trait_resource_listing_head_200'
+        "404":
+          description: The requested Source does not exist.
     get:
       summary: Source Tags
       description: Returns the Source tags.
@@ -405,6 +407,8 @@ paths:
           description: No content. The tag has been updated.
         "400":
           description: Bad request. Invalid Source tag value.
+        "404":
+          description: The requested Source does not exist, or the tag name in the path is invalid.
     delete:
       summary: Delete Source tag
       description: Delete a specific tag on a Source
@@ -414,6 +418,8 @@ paths:
       responses:
         "204":
           description: No content. The Source segment has been deleted.
+        "404":
+          description: The requested Source ID or tag in the path is invalid.
   /sources/{sourceId}/description:
     parameters:
       - name: sourceId
@@ -469,6 +475,8 @@ paths:
           description: No content. The description has been updated.
         "400":
           description: Bad request. Invalid Source description.
+        "404":
+          description: The requested Source does not exist.
     delete:
       summary: Delete Source Description
       description: Delete the description property.
@@ -478,6 +486,8 @@ paths:
       responses:
         "204":
           description: No content. The Source description property has been deleted.
+        "404":
+          description: The Source ID in the path is invalid.
   /sources/{sourceId}/label:
     parameters:
       - name: sourceId
@@ -533,6 +543,8 @@ paths:
           description: No content. The label has been updated.
         "400":
           description: Bad request. Invalid Source label.
+        "404":
+          description: The requested Source does not exist.
     delete:
       summary: Delete Source Label
       description: Delete the label property.
@@ -542,6 +554,8 @@ paths:
       responses:
         "204":
           description: No content. The Source label property has been deleted.
+        "404":
+          description: The requested Source ID in the path is invalid.
   /flows:
     head:
       description: Return flows path headers
@@ -757,6 +771,8 @@ paths:
           description: Bad request. Invalid flow JSON.
         "403":
           description: Forbidden. You do not have permission to modify this flow. It may be marked read-only.
+        "404":
+          description: The requested Flow ID in the path is invalid.
     delete:
       description: |
         Deletes the flow and associated segments. If flow segment deletion
@@ -783,6 +799,8 @@ paths:
           description: No content. The flow has been deleted and the flow segments have been or will be deleted.
         "403":
           description: Forbidden. You do not have permission to modify this flow. It may be marked read-only.
+        "404":
+          description: The requested Flow ID in the path is invalid.
   /flows/{flowId}/tags:
     parameters:
       - name: flowId
@@ -799,6 +817,8 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/trait_resource_listing_head_200'
+        "404":
+          description: The requested flow does not exist.
     get:
       description: Returns the flow tags.
       operationId: GET_flows-flowId-tags
@@ -871,6 +891,8 @@ paths:
           description: Bad request. Invalid flow tag value.
         "403":
           description: Forbidden. You do not have permission to modify this flow. It may be marked read-only.
+        "404":
+          description: The requested flow does not exist.
     delete:
       description: Delete the tag.
       operationId: DELETE_flows-flowId-tags-name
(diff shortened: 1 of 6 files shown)


api/TimeAddressableMediaStore.yaml
```diff
@@ -1019,6 +1043,8 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/trait_resource_listing_head_200'
+        "404":
+          description: The requested flow does not exist.
```
philipnbbc, line 1047: Should this be a 404 if the flow ID path component is invalid and not when the Flow does not exist? If a Flow does not exist it returns `[]`.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1117,6 +1143,10 @@ paths:
                     expressed a preference for the "pipeline-a" URLs by putting them first in the list.
                   value:
                     $ref: examples/flow-segments-get-200-multiple-urls.json
+        "400":
+          description: Bad request. Invalid query options.
+        "404":
+          description: The requested flow does not exist.
```
philipnbbc, line 1149: Same comment as above for the HEAD request.

api/schemas/flow-audio.json
```diff
@@ -33,15 +33,18 @@
...
             "channels": {
               "description": "The channel count.",
-              "type": "integer"
+              "type": "integer",
+              "exclusiveMinimum": 0
             },
             "bit_depth": {
-              "description": "The number of significant bits used to represent the audio sample. The minumum number of bytes then equals `round_up(bit_depth / 8)`.",
-              "type": "integer"
+              "description": "The number of significant bits used to represent the audio sample. The minumum number of bytes then equals `round_up(bit_depth / 8)`. If codec is `audio/x-raw-int` or `audio/x-raw-float`, bit_depth must be set.",
```
philipnbbc, line 45: Slight improvement regarding float vs double:
```suggestion
              "description": "The number of significant bits used to represent the audio sample. The minumum number of bytes then equals `round_up(bit_depth / 8)`. If codec is `audio/x-raw-int` bit_depth must be set. If codec is `audio/x-raw-float` bit_depth must be set to 32 or 64",
```

api/schemas/flow-core.json
```diff
@@ -75,11 +76,13 @@
     },
     "avg_bit_rate": {
       "description": "The average bit rate of the flow content in 1000 bits/second. A precise definition of how the bit rate is calculated is not provided here.",
-      "type": "integer"
+      "type": "integer",
+      "exclusiveMinimum": 0
```
philipnbbc, line 80: I could foresee an average and maximum bit rate of 0 for Flows that don't yet have content or below 1kb/s.

api/schemas/flow-core.json
```diff
@@ -75,11 +76,13 @@
...
     },
     "max_bit_rate": {
       "description": "The maximum bit rate of the flow content in 1000 bits/second. A precise definition of how the bit rate is calculated is not provided here.",
-      "type": "integer"
+      "type": "integer",
+      "exclusiveMinimum": 0
```
philipnbbc, line 85: See comment for `avg_bit_rate`.

api/schemas/flow-video.json
```diff
@@ -40,7 +40,8 @@
                             "properties": {
                                 "numerator": {
                                     "description": "numerator",
-                                    "type": "integer"
+                                    "type": "integer",
+                                    "exclusiveMinimum": 0
```
philipnbbc, line 44: Might be useful to add a clarification to the description that if `frame_rate` is not provided then either it is unknown or it is known to be variable frame rate. Not sure, but we could think about whether a zero frame rate is used to indicate a variable frame rate rather than lump it in with unknown?

philipnbbc, line 44: Nit: add `"exclusiveMinumum": 0` to the frame rate denominator as well

api/schemas/flow-video.json
```diff
@@ -137,12 +143,16 @@
                             ]
                         },
                         "horiz_chroma_subs": {
-                            "description": "Horizontal chroma component sub-sampling.",
-                            "type": "integer"
+                            "description": "Horizontal chroma component sub-sampling. When unc_type is set to a YUV type, horiz_chroma_subs must be set.",
+                            "type": "integer",
+                            "exclusiveMinimum": 0,
+                            "maximum": 2
```
philipnbbc, line 149: Not sure whether this limit needs or should be set unless we are restricting the formats supported. Wikipedia for example [lists 4:1:0 and 3:1:1](https://en.wikipedia.org/wiki/Chroma_subsampling)

api/schemas/flow-video.json
```diff
@@ -137,12 +143,16 @@
...
                         },
                         "vert_chroma_subs": {
-                            "description": "Vertical chroma component sub-sampling.",
-                            "type": "integer"
+                            "description": "Vertical chroma component sub-sampling. When unc_type is set to a YUV type, vert_chroma_subs must be set.",
+                            "type": "integer",
+                            "exclusiveMinimum": 0,
+                            "maximum": 2
```
philipnbbc, line 155: See comment above for `horiz_chroma_subs`

api/TimeAddressableMediaStore.yaml
```diff
@@ -657,8 +669,6 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/trait_resource_info_head_200'
-        "400":
```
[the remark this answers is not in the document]
j616: The HEAD doesn't list any query options. I guess it should list the same ones as the GET?

api/schemas/flow-core.json
j616: Would this be correct usage? Arguably this parameter shouldn't be set at all until the content becomes available? I guess some infrequent data  flows may have extremely low data rates, though.

api/schemas/flow-video.json
j616: I wonder if we would be better signalling it explicitly through a separate parameter? Either way, we should probably wait until we have more concrete use cases before making that decision.

api/schemas/flow-video.json
j616: Fair. I'll remove the max and set the minimum to inclusive 0.

api/TimeAddressableMediaStore.yaml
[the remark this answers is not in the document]
philipnbbc: I think so yes.

api/schemas/flow-core.json
philipnbbc: I don't think a bit rate value 0 would cause issues for a process reading the content. Ideally if the bit rate is unknown, because there is no content or it is something that couldn't be parsed or calculated, then it is omitted. A TAMS store could also decide to set it to 0 rather than deleting the property after deleting all segments.

api/schemas/flow-video.json
philipnbbc: I think the minimum should be 1, i.e. no-sub-sampling.

j616: Bah. 4:2:0 confusion strikes again

api/TimeAddressableMediaStore.yaml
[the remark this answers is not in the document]
j616: Added missing HEAD parameters in 682b86fe77923ab1d5ae94abeea7411636b2ab21

philipnbbc (approved):
LGTM

PR #71, opened 2024-07-02
gha: trigger CI workflow after release to update API docs

philipnbbc:
# Details
The `release` trigger didn't work because the release used the GITHUB_TOKEN and that will not create a new workflow run - see [Triggering a workflow from a workflow](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow)

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187836655

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -10,6 +10,11 @@ on:
     branches:
       - 'main'
 
+  workflow_run:
+    workflows: [Release Workflow]
+    types:
+      - completed
+
   workflow_dispatch:
     inputs:
       forceDocsUpload:


j616 (approved):
LGTM. Though I'm not sure the square brackets are needed for a single workflow. See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-a-single-event

philipnbbc:
> LGTM. Though I'm not sure the square brackets are needed for a single workflow. See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-a-single-event

All the examples for [workflow_run](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) uses YAML list notation (square brackets or '-' child)  for the `workflows` property and it doesn't mention a single workflow shortcut. I think then it's probably better to follow the examples.

PR #75, opened 2024-07-22
api: allow flow segment paging using opaque key

philipnbbc:
# Details
Allow the next page link to include a "page" query parameter that allows the next page to be retrieved. If the TAMS implementations uses the "page" parameter then it will likely leave the "timerange" query parameter unchanged.

Marking this change as a feature rather than a breaking change because clients should be treating the query parameters of a page link as opaque (given that the "page" parameter is opaque and any potential relationship with the other parameters is unknown).

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/187930075

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1179,13 +1179,18 @@ paths:
           description: Filter on object identifier.
           schema:
             type: string
+        - name: timerange
+          in: query
+          description: Return only the results in the timerange specified.
+          schema:
+            $ref: 'schemas/timerange.json'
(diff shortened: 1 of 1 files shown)


j616 (approved):
LGTM. As mentioned, we seem to be missing headers on HEAD requests for pages endpoints. I've added a ticket for this.

PR #87, opened 2024-09-09
Application note for working with C2PA provenance in TAMS

chalford:
# Details
Add an application note detailing an approach for working with C2PA provenance data in TAMS Flows.

# Pivotal Story (if relevant)
Story URL: N/A

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -50,4 +50,4 @@ For more information on how we use application notes, see [here](./appnotes/READ
 | [0008](./appnotes/0008-timestamps-in-TAMS.md)                    | Timestamps in TAMS                                            |
 | [0009](./appnotes/0009-storage-label-format.md)                  | Storage label format specification                            |
 | [0010](./appnotes/0010-long-running-sources-and-flows.md)        | Long-running Sources and Flows                                |
-| [0010](./appnotes/0011-c2pa.md)        | C2PA provenance across related Sources and Flows |
+| [0011](./appnotes/0011-c2pa.md)                                  | C2PA provenance across related Sources and Flows |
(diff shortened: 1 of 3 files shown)


docs/README.md
```diff
@@ -50,3 +50,4 @@ For more information on how we use application notes, see [here](./appnotes/READ
 | [0008](./appnotes/0008-timestamps-in-TAMS.md)                    | Timestamps in TAMS                                            |
 | [0009](./appnotes/0009-storage-label-format.md)                  | Storage label format specification                            |
 | [0010](./appnotes/0010-long-running-sources-and-flows.md)        | Long-running Sources and Flows                                |
+| [0010](./appnotes/0010-long-running-sources-and-flows.md)        | C2PA provenance across related Sources and Flows
```
j616, line 53: Could you update the App Note number and link here?

j616:
Looks like this isn't currently passing the markdown linter. You should be able to run this locally with `make lint` if you have docker installed

chalford:
> Looks like this isn't currently passing the markdown linter. You should be able to run this locally with `make lint` if you have docker installed

Thanks. The remaining errors are:

* Inline HTML `a` element. This was to allow in-doc heading links/references. Is there a better way to do this, or should I replace with some text (e.g. see "Heading")
* Inline HTML center element. This was for the diagram captions. Markdown doesn't seem to have a solution here. Shall I just remove and let the caption left align?

j616 (approved):
Changes look good. Made a trivial change myself to fix a missed instance of the document number

PR #90, opened 2024-11-04
ADR: Add Source edit by reference endpoints

samdbmg:
# Details
Adds an ADR proposing how a Source-level edit-by-reference endpoint might behave, along with a partial implementation to illustrate one of the options.

Draft for discussion.

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/188413555

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

- [x] PR added to Pivotal story (if relevant)
- [x] Follow-up stories added to Pivotal

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,274 @@
...
+        "timerange": "[0:0_9:0)",
+        "ts_offset": "0:0",
+    },
+    {
+        "reference": {
+            "flow_id": "flow-d-id",
+            "timerange": "[0:0_2:0)"
+        },
+        "timerange": "[9:0_11:0)",
+        "ts_offset": "0:0",
```
philipnbbc, line 137:
```suggestion
        "ts_offset": "9:0",
```

docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,274 @@
...
+
+As above, however store implementations cannot "invent" the relevant segments: they merely provide detail of the Source timeline and leave the rest to the client.
+
+This option has the same list of pros and cons as Option 2, however the following benefit is removed:
+> Good, because at read time the references are transparent to a client: the store can read-through them.
+
+And the following drawback is added:
+> Bad, because clients have to do significant additional work to make use of a Source timeline (identifying, de-referencing and re-mapping the relevant Flows).
+
+### Option 3: Provide additional Flow Segment API capability for more direct by-reference operations
```
philipnbbc, line 114: Some general questions and thoughts on this option:
* I think in addition to using parts (time bounded) of a Flow in another Flow that defining a new Flow that is a time shift of another Flow would be useful.
* Could the reference Flow Segment be unbounded? This would be useful in a live context and allow time shifting without resorting to time bounding.
* The Flow Segment reference timerange may fall within media objects. This means that more information is required in a dereferenced Flow Segment to allow a client to work out which is the first or last sample in the target timerange.
* What limitations if any should be placed on reference Flow Segments and what implementation recommendations should be given? E.g. multi levels of references to other reference Flow Segments. Are limits an implementation choice?

docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,274 @@
...
+            },
+            "available_image_bounds": null,
+            "target_url": "tamss://tams.example.com/flows/9bb414a5-862c-494f-86ce-8e2720ecc315"
+        }
+    },
+    "active_media_reference_key": "DEFAULT_MEDIA"
+}
+```
+
+### Reference Form
```
philipnbbc, line 240: This extends the URL form with metadata? Or should the `target_url` be changed to a TAMS base URL?

philipnbbc (approved):
I like the choices of 3 and 4 but think that some implementation investigation would be needed to fully understand the details and consequences of 3.

docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,274 @@
...
+An additional endpoint may be added in future to directly query where an object ID is used, and may be the subject of a future ADR.
+Regardless, there is no way to reason about how a given Flow came in to existence: was it the originator of these segments, a copy or a copy-of-a-copy - for some applications (e.g. rights management) this can be quite important.
+
+This ADR discusses some options for improving on these limitations.
+
+## Considered Options
+
+* Option 1: Provide an edit API in the store that allows more complex operations to be specified on Sources
+* Option 2: Provide a limited API for simple cut operations on Sources
+* Option 2a: Provide a limited API as in (Option 2), that prevents mixing Flow and Source operations
```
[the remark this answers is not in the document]
samdbmg: Oops - fixed in 1b305cf

docs/adr/0024-source-level-edit.md
samdbmg: Good point - fixed in 1b305cf

docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,274 @@
...
+### Option 2a: Provide a limited API as in (Option 2), that prevents mixing Flow and Source operations
+
+As above, however if the Source-level edit endpoint is used on a given Source, Flow Segments cannot be created for the Flows.
+Instead the Source, and all the Flows that represent it, must be fully described using references to other Sources.
+In the case where new segments need to be written (e.g. to cover a transition) a new Flow and Source can be created as the "transition layer", which can then be composed into a new Source.
+
+Building on the example above, instead we now have Flow/Source pairs A, B and D, where D contains only the rendered dissolve between A and B.
+The resulting Flow/Source pair C contains `SourceA@[0:0_9:0)`, `SourceD@[9:0_11:0)`, `SourceB@[1:0_10:0)`.
+
+This option has the same list of pros and cons above, expect the following item is mitigated and removed:
```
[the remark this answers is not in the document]
samdbmg: Good spot - fixed in 1b305cf

docs/adr/0024-source-level-edit.md
samdbmg: I've made this a `MissingReference` in 1b305cf and expanded the text to better explain why the `metadata` is also useful

docs/adr/0024-source-level-edit.md
samdbmg: > - I think in addition to using parts (time bounded) of a Flow in another Flow that defining a new Flow that is a time shift of another Flow would be useful.
> - Could the reference Flow Segment be unbounded? This would be useful in a live context and allow time shifting without resorting to time bounding.
Perhaps the requirement should be "bounded before" so you can time-shift but still calculate the necessary offset? The other ways would be to add a form that has `reference.timeshift` instead of `reference.timerange` or similar, or require "bounded one end", but an unbound timeshift going backwards in time feels like an unusual requirement.

> The Flow Segment reference timerange may fall within media objects. This means that more information is required in a dereferenced Flow Segment to allow a client to work out which is the first or last sample in the target timerange.
A way around this would be to treat `reference.timerange` like the `timerange` query parameter on `/flows/<flowid>/segments`: it selects all Flow Segments where any part falls within the range. That would probably mean the start of whatever got selected is moved to the start of the `timerange` given (so if you set `reference.timerange` to the middle of a segment the resulting shift is later than you wanted) which leads to wording to the effect that "clients SHOULD ensure the start of a reference timerange aligns with the start of a segment in the referenced Flow".
The other way is to add more info to the dereferenced Flow Segment - e.g. setting `sample_offset` to come out at the right point within the media object, or adding another field that sub-selects within a media object TimeRange.

> What limitations if any should be placed on reference Flow Segments and what implementation recommendations should be given? E.g. multi levels of references to other reference Flow Segments. Are limits an implementation choice?
Ideally I think references-to-references get accepted in a POST and a store implementation figures out how to deference that efficiently, however there has to be a limit on that (and some mechanism to handle infinite reference loops).

We could say that "a store implementation MAY store and return a different (but functionally equivalent) list of Flow Segments" or something, so you can resolve nested references on write. You might write a reference to Flow X but Flow X is a reference to Flows Y and Z, so a GET (with `?deference=false`) returns a reference to Flows Y and Z.

Reference loops are harder, because ideally we'd find them on write but that might create quite a complex query to explore the reference tree in both directions, although maybe we have to do that anyway to make reference counting work.

philipnbbc: > > * I think in addition to using parts (time bounded) of a Flow in another Flow that defining a new Flow that is a time shift of another Flow would be useful.
> > * Could the reference Flow Segment be unbounded? This would be useful in a live context and allow time shifting without resorting to time bounding.

>  Perhaps the requirement should be "bounded before" so you can time-shift but still calculate the necessary offset? The other ways would be to add a form that has `reference.timeshift` instead of `reference.timerange` or similar, or require "bounded one end", but an unbound timeshift going backwards in time feels like an unusual requirement.

Timeshift for negative infinity would for example be needed if the Flow content is being populated in a random (asynchronous) order and you may not know or care where it starts. It would be nice to be able to simply define Flow A to be a timeshift of Flow B no matter what the bounds may be of Flow B. If there is a bound in the `reference.timerange` (or consequently in `timerange`) then the timeshift can be calculated from that and otherwise a `reference.timeshift` would be needed.

My current thinking is that if we allow any unboundedness in the reference Flow Segment then allowing fully unbounded seems sensible and in that case a `reference.timeshift` is needed if a timeshift operation is supported.


> > The Flow Segment reference timerange may fall within media objects. This means that more information is required in a dereferenced Flow Segment to allow a client to work out which is the first or last sample in the target timerange.

> A way around this would be to treat `reference.timerange` like the `timerange` query parameter on `/flows/<flowid>/segments`: it selects all Flow Segments where any part falls within the range. That would probably mean the start of whatever got selected is moved to the start of the `timerange` given (so if you set `reference.timerange` to the middle of a segment the resulting shift is later than you wanted) which leads to wording to the effect that "clients SHOULD ensure the start of a reference timerange aligns with the start of a segment in the referenced Flow".
> The other way is to add more info to the dereferenced Flow Segment - e.g. setting `sample_offset` to come out at the right point within the media object, or adding another field that sub-selects within a media object TimeRange.

I agree with you that a way around this would be to have clients treat the reference Flow Segment's `timerange` and `reference.timerange` as a query parameter and it's the client's problem to handle the case where the timerange falls within a media object. I guess adding a SHOULD that it should fall at the start and end of media objects is something that could be stated if we want to keep things simple for clients. The downside is that it might hamper reuse because the content needs to be ideally segmented.

I'm leaning towards not requiring `sample_offset` etc. in a reference Flow Segment because it would require the creator to download the segment and work out what the values should be. It is a choice between the creator of the reference Flow Segment figuring out what the `sample_offset` and/or `sample_count` should be for the first and last dereferenced Flow Segment, or leave it to the reader of the Flow Segment to work it out like they would need to do if running  a query with timerange.

In a regular Flow Segment the `timerange` is for the whole media file unless `sample_start` and/or `sample_count` are set in which case the `timerange` is for the clipped portion of the media file. Maybe what should be added is a `media_object_timerange` for the whole media object which then allows a reference Flow Segment to be dereferenced to a Flow Segment that has `timerange` != `media_object_timerange` and `sample_offset` and/or `sample_count` are missing. A client would then need to work out the sample offset and count using the 2 timeranges given.


> > What limitations if any should be placed on reference Flow Segments and what implementation recommendations should be given? E.g. multi levels of references to other reference Flow Segments. Are limits an implementation choice?

> Ideally I think references-to-references get accepted in a POST and a store implementation figures out how to deference that efficiently, however there has to be a limit on that (and some mechanism to handle infinite reference loops).
> 
> We could say that "a store implementation MAY store and return a different (but functionally equivalent) list of Flow Segments" or something, so you can resolve nested references on write. You might write a reference to Flow X but Flow X is a reference to Flows Y and Z, so a GET (with `?deference=false`) returns a reference to Flows Y and Z.
> 
> Reference loops are harder, because ideally we'd find them on write but that might create quite a complex query to explore the reference tree in both directions, although maybe we have to do that anyway to make reference counting work.

Agree. Maybe another limitation a store could optionally apply is whether `reference.timerange` starts and end on the non-reference Flow Segment `timerange`, possibly also whether a non-reference Flow Segment does not cover the media object timerange and that `sample_offset` / `sample_count` are missing.

docs/adr/0024-source-level-edit.md
```diff
@@ -0,0 +1,278 @@
...
+        "reference": {
+            "flow_id": "flow-c-id",
+            "timerange": "[1:0_10:0)"
+        },
+        "timerange": "[11:0_20:0]",
+        "ts_offset": "10:0",
+    }
+]
+```
+
```
himslm01, line 150: Just checking - am I correct in my understanding that the References could be pointing to Timeranges of Flows which have no segments associated, and that would be just fine?

That could mean that, for instance, a Reference could point to a Timerange of a Flow which is "in the future" and therefore may have new segments added.

Thus a "time shifted" Flow could be created from a Timerange of a Flow such as `now()` until an infinite future time such that a, as-yet non-existent, future-playout-device could play out a Flow which is "still in record"?

samdbmg: That's my intention, yes, although see the bottom of https://github.com/bbc/tams/pull/90#discussion_r1907135373 for some of the complications that creates regarding things like referential loops (and a discussion about whether a more explicit `timeshift` feature is needed)

docs/adr/0024-source-level-edit.md
thomasheritage: For Option 3 it does sound like you could consider making it "time-shifts only" rather than going as far as "cuts only" (which is of course already a scaling back on being able to describe a "full composition").
If time-shifting is the main requirement then that avoids getting into any of the complications that allowing cuts would introduce...

samdbmg:
This ADR was discussed at some length both within the BBC R&D TAMS team, and with the BBC B&EUT architects and AWS Solution Architects, which led to a clearer path forward.

## BBC R&D TAMS Team - 15th January 2025
- We discussed whether only allowing cuts in a simple reference is sufficient (versus being able to fully describe composition), and concluded there is value in having both, but the need for clarity about the distinctions
  - For example, expressing that there is both a mechanism to refer to TAMS content in something like OpenTimelineIO, and also a way to write that back as a Flow that doesn't duplicate essence
- Similarly, we considered whether an EDL format such as OpenTimelineIO is sufficient without any expansion to the TAMS API, and concluded a simple reference model is also useful for other use cases 
- We considered whether OpenTimelineIO references should primarily be Sources, and concluded that made sense because editorial changes should be described in terms of Sources
  - However there is a need for a mechanism to map Sources onto a preferred Flow (there are ongoing discussions about the concept of a `profile`, but no writeup yet).
- In addition, the need for an approach to describe references to remote stores was identified, and will be the subject of a future AppNote

## BBC R&D/BBC B&EUT Architects/AWS Solution Architects - 17th January 2025
- Practical use cases were discussed in the BBC's News production workflows, and how this would simplify some aspects
  - Being able to pass a "sub-clip" from a long recording (e.g. an agency feed) into another system using TAMS is useful, rather than having to pass the entire multi-day recording
  - As is being able to pass that Source in a way that it continues to grow and allowing an editor to view it in the usual way: for example Prime Minister's Questions happens at a fixed time and could be sub-clipped to pass to an editor
- Using some sort of `profile` field to allow Source-based referencing (Option 2) and then map that down to Flows was considered
  - However it would require adding an entire new timeline mechanism on Sources that doesn't currently exist
  - It is also still limited to only handling very simple composition operations (cuts between other Sources)
  - Furthermore it is difficult to manage how Flows come into existence in all but the simplest cases where an exact match is already present
- Circular references in Option 3 were explored, concluding that some care be taken that the proposed mechanism did not allow a user to disrupt other content or operations, by maliciously creating a circular reference
- The addition of an endpoint for bulk creation of Flow Segments was proposed

This topic is sufficiently complex it may be wise to split it up into some additional ADRs

samdbmg:
This PR has now been tidied up with the conclusions from various discussions, and a followup (#104) opened with some Application Notes that follow on from it.
As far as I'm concerned, it's now largely ready to merge, and the remaining areas (precise details of Flow References in Option 3, and bulk edit of segments) will form future ADRs that are tracked in our planning system.
I'll leave this one open for a few more days for any final comments

PR #86, opened 2024-08-28
adr: add flow bit rate property definitions ADR

philipnbbc:
# Details
This PR adds draft ADR for discussion. It documents possible Flow bit rate property definitions and additional related properties.

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/188127666

# Related PRs
_Where appropriate. Indicate order to be merged._

# Submitter PR Checks
_(tick as appropriate)_

- [ ] PR completes task/fixes bug
- [ ] API version has been incremented if necessary
- [ ] ADR status has been updated, and ADR implementation has been recorded
- [ ] Documentation updated (README, etc.)
- [ ] PR added to Pivotal story (if relevant)
- [ ] Follow-up stories added to Pivotal

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] PR completes task/fixes bug
- [ ] Design makes sense, and fits with our current code base
- [ ] Code is easy to follow
- [ ] PR size is sensible
- [ ] Commit history is sensible and tidy

# Info on PRs
The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.

PR #91, opened 2024-11-13
Add flow property updates ADR and specification extensions

philipnbbc:
# Details
This PR came about because an question was raised about whether the Flow `source_id` can be updated. The proposal is to allow it to be updated to fix incorrect values for example. This may be restricted by systems that sit in front of TAMS API that check permissions for example.

This PR offers some changes wrt updating Flow properties but proposes no change at this stage. Requirements may arise in future that changes this.

This PR adds endpoints for bit rate (see also https://github.com/bbc/tams/pull/86) and flow collection properties. This is to make it consistent with other properties that are likely to be updated.

# Pivotal Story (if relevant)
Story URL: https://www.pivotaltracker.com/story/show/188210268

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

- [ ] Follow-up stories added to git push --set-upstream origin philipn-flow-property-updatesPivotal

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/adr/0025-flow-property-updates.md
```diff
@@ -0,0 +1,248 @@
...
+* Option c2: Allow a PATCH update of a Flow
+* Option c3: Add endpoints for PUTing to defined groups of properties
+* Option c4: Continue to use the current Flow PUT and property updates
+
+## Decision Outcome
+
+The chosen options are: a2, b3 and c4.
+
+Option a2 is chosen to allow all properties (with some exceptions) to be corrected and not require a Flow to be complete on creation.
+It is expected that systems that sit in front of the TAMS API apply additional constraints on updates for a particular installation.
```
samdbmg, line 137: There's a flexibility vs data integrity thing to think about I suppose: it would be frustrating if a client came to depend on the "adjust the Source ID after creation" mechanic only to run across an implementation that doesn't let you do that. Is there a compromise along the lines of "clients SHOULD NOT change these properties after creation, however implementations MAY permit it to allow operators to correct mistakes and simplify client implementations" or something?

samdbmg (approved):
LGTM - barring the inlined comment I agree with the decisions too

api/examples/flow-avg-bit-rate-get-200.json
```diff
@@ -0,0 +1 @@
+3246
```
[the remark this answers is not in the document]
philipnbbc: Agree, it's simpler to have these single value examples in the spec. I've moved the ones added in this PR as well as the Flow `read_only` examples.

docs/adr/0025-flow-property-updates.md
philipnbbc: Makes sense. I've added similar words to the ADR.

PR #102, opened 2025-01-23
Flag that instantaneous TimeRanges can't use exclusive markers

j616:
# Details
Flag that instantaneous TimeRanges can't use exclusive markers

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5375

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


philipnbbc (approved):
I would say something equivalent to "cannot" rather than "should not" because a single timestamp with exclusive markers is invalid and should be treated as an empty / "never" timerange. "should not" suggests there are exceptional cases where it can be used for instantaneous timestamps. The [TimeRange representation](https://github.com/bbc/tams/blob/main/docs/appnotes/0008-timestamps-in-TAMS.md#timerange-representation) description has the point "A TimeRange should be treated as an empty TimeRange by implementations if the end is before the start, or the start and end are equal and either has an exclusive marker."

Otherwise LGTM

edit: added " used for instantaneous timestamps" to clarify that if it has exclusive markers it is not an instantaneous timestamp by definition.

PR #103, opened 2025-01-24
Fix commit check checkout ref

philipnbbc:
# Details
Using `github.ref` resulted in `fetch-depth` number of commits from the `main` branch. This resulted in a commit check failure because of an earlier [commit](https://github.com/bbc/tams/commit/0e6712106a219491d0a4f203f54c3bb21ff95e98) that had the word SQUASH in it.

# Jira Issue (if relevant)
Jira URL:

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/.github/workflows/commit-check-workflow.yml
+++ b/.github/workflows/commit-check-workflow.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: ${{github.event.pull_request.commits}} # only checkout commits from this PR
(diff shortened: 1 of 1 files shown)


samdbmg (approved):
LGTM - while we can't do this for the main build workflow (because we need to build the result of the merge) it's safe to do for the commit check

PR #104, opened 2025-01-27
Add Appnotes describing external references and using OpenTimelineIO

samdbmg:
# Details
- Adds an appnote with suggested reference formats for referring to TAMS content in other systems 
- Adds an appnote describing how to use TAMS references in OpenTimelineIO compositions 

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-3522

# Related PRs
Builds on and implements parts of #90 

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/appnotes/0015-using-tams-in-opentimelineio.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+Note the presence of an `available_range_offset` that describes how the `available_range` here maps onto the Flow timeline, much as `ts_offset` remaps media essence timing to Flow timing.
+
+### ExternalReference Format
+
+This format refers to a URL in a TAMS instance and points to a specific, concrete Flow.
+It is possible a Clip could have more than one `ExternalReference` object (e.g. for each Flow), with the appropriate default selected: this allows, for example, online and proxy workflows to be presented in a UI.
+
+Notice that the URL has a prefix `tamss://` (for "TAMS Secure" - `tams://` would also work for HTTP).
```
philipnbbc, line 83: I wonder whether it is simpler to just have a `tams://` pseudo-protocol. If secure/non-secure is an option then I don't see why this `tams` locator needs to decide either way - that's for the linker to decide when translating to the actual protocol.

Is it worth stating why you would use `tams://` rather than `https://`, given that they look essentially the same?

Given my comments on 0014, one option would be to separate out the TAMS location (e.g. the base url). This might make it easier for linkers that want or need to use a different location to have the identity part separate from the (assumed / default) location. However, it's just a thought and I'm not advocating changing the design here.

docs/appnotes/0014-referencing-tams-content-in-other-systems.md
```diff
@@ -0,0 +1,78 @@
...
+When referring to a TAMS object as a string, the URI reference style above may be used, and will make sense in many cases.
+
+If a clear need arises, alternative forms such as URNs or [tag URIs](https://www.rfc-editor.org/rfc/rfc4151.html) will be added via the ADR process.
+
+## Live Signal Domain
+
+The TAMS identity model is deliberately aligned with that of [AMWA NMOS MS-04](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/2.1._Summary_and_Definitions.html), to facilitate alignment of fast-turnaround, live and file-based workflows.
+It is expected that when content is ingested into TAMS in an NMOS environment, the Flow and Source IDs and internal timing of that content is preserved and can be considered as part of a unified control system.
+
+When content is played out into a live signal environment, it is likely to be time-shifted to align with other content, which requires that a new Source and Flow ID be assigned (as described in the [MS-04 Playout Server example](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.2._Composite_Media_Operations.html#playout-server)).
```
philipnbbc, line 74: This is a good use case for having the lightweight copy option.

docs/appnotes/0015-using-tams-in-opentimelineio.md
```diff
@@ -0,0 +1,171 @@
...
+
+![Input material and basic composition](./images/0015-opentimeline-writeback-composition.png)
+
+Clearly the result of rendering this timeline is a new Source and Flow, and it could be rendered in the usual way, producing new essence that is written to the store.
+However, it could also be rendered by identifying areas of the timeline that copy the inputs directly, and creating Flow references instead.
+In the diagram below Source/Flow `sO/fO` have been created as the output, containing a mix of references to inputs, and new material where references are not possible.
+
+![Render process and new Flows](./images/0015-opentimeline-writeback-render.png)
+
+This makes the render process faster (references are a very "cheap" metadata operation), makes more efficient use of storage, and avoids an additional decode-encode cycle on some of the content.
```
philipnbbc, line 150: (repeating discussions that have been had and the description further below)
and the reason for having those references in TAMS rather than keeping it in an external EDL is that it adds the bits that would otherwise to missing from the Flow that contains the new material segments - it becomes more a published artefact rather than an ephemeral work-in-progress artefact.. It can then be played back by processes that may not have the capabilities to render (parts of) the EDL. Although the other option is to pass a flattened EDL that contains the missing references to other Flows.

philipnbbc (approved):
Typos, some commentary (not requiring changes) and question inline. Otherwise LGTM

docs/appnotes/0014-referencing-tams-content-in-other-systems.md
```diff
@@ -0,0 +1,78 @@
+# 0014: Referencing TAMS content in other systemss
+
+## Abstract
+
+Many workflows require that TAMS content be referenced in other systems or operations, or example linking it to an asset in a MAM, using it to describe an edit composition or integrating it with signals in the live domain.
+This Application Note describes a number of common reference formats to describe these use cases.
+
+This is intended to be a working document, and as new use cases are built out, new formats may be added.
+
+### URI References
```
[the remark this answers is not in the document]
samdbmg: I've added a few words to the effect that you can use a URI (since there has to be a TAMS store _somewhere_) but ignore the location parts if you want to.

Beyond that is String references I think: given that RFC8141 [deprecates](https://datatracker.ietf.org/doc/html/rfc8141#appendix-C) the `urn:x-` style (and registering a formal namespace seems like quite the scope creep!) I'm not sure URNs are the way to go - another approach is [RFC4151 tag URIs](https://www.rfc-editor.org/rfc/rfc4151.html) which would suggest something like `tag:bbc.github.io,2025:tams:flow:f8a64ec6-f5d4-43ba-90b4-3dfa317d133b`

docs/appnotes/0014-referencing-tams-content-in-other-systems.md
```diff
@@ -0,0 +1,78 @@
...
+
+However there is no direct API concept to refer to a specific TimeRange of a Source.
+Instead, this form can be used, and a client can map it accordingly after traversing the Source/Flow graph.
+
+```text
+https://tams.example.com/sources/7abe4d59-7d1c-432d-ba2a-6180e12137cc?timerange=[100:0_200:0)
+``
+```
+
+## JSON Object References
```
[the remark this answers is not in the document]
samdbmg: Yes, including some way to identify the type of thing would make sense, but I suspect we're at the point where the ensuing API break isn't worth it

docs/appnotes/0015-using-tams-in-opentimelineio.md
samdbmg: Good to call this out - I've added a few words to this effect

docs/appnotes/0014-referencing-tams-content-in-other-systems.md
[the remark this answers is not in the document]
samdbmg: I've opened https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5402 to track considering this

PR #108, opened 2025-02-03
example: Add interval edit script

samdbmg:
# Details
Adds a script that cuts between Flows on a fixed time interval, demonstrating `sample_offset` and `sample_count`.
Also updates the `simple_edit` script to set some custom tags on created Flows.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5392

# Related PRs
Follows #107 (because you can't test this script without those fixes)

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/examples/simple_edit.py
+++ b/examples/simple_edit.py
@@ -72,6 +72,7 @@ async def get_segments(
     flow_id: UUID,
     timerange: TimeRange
 ) -> list[dict]:
+    """Fetch a single page of segments from the given Flow"""
     async with get_request(
         session,
         credentials,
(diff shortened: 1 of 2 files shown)


examples/simple_edit.py
```diff
@@ -152,6 +175,113 @@ async def simple_edit(
...
+            # Draw a segment from the current list (and drop segments if we've passed them)
+            next_seg = current_seg["list"][0]
+            next_seg_tr = TimeRange.from_str(next_seg["timerange"]).normalise(edit_rate.numerator,
+                                                                              edit_rate.denominator)
+            if next_seg_tr.end <= position_in_flow_timeline:
+                print(f"Segment {next_seg_tr} is before current position {position_in_flow_timeline} - dropping")
+                current_seg["list"].popleft()
+                continue
+
+            next_seg_offset = Timestamp.from_str(next_seg.get("ts_offset", "0:0"))
```
philipnbbc, line 235: I was slightly confused about what each offset was from the name used. Could this one (which is only used once) be called `next_seg_ts_offset` and rename the `offset` in `current_seg` and `other_seg` to `timeshift`? However, it's a personal choice and so I leave it to you to decide what makes sense.

examples/simple_edit.py
```diff
@@ -152,6 +175,113 @@ async def simple_edit(
...
+
+            # Work out how many seconds into the underlying segment we start
+            seg_start_offset = (working_time + current_seg["offset"]) - next_seg_tr.start
+
+            # Work out how much of the segment to include in the output
+            segment_length_remaining = next_seg_tr.end - position_in_flow_timeline
+            remaining_time_in_cut = next_switch_at - working_time
+
+            if (remaining_time_in_cut < segment_length_remaining):
```
philipnbbc, line 244: There will be some anomalies when doing this and other comparisons if the media rates are not integers (because of rounding errors in some of the calculations using nanoseconds), but I think on balance it isn't helpful to have those complications in this example script.

philipnbbc (approved):
A couple comments inline (and a bug found elsewhere) and otherwise LGTM

examples/simple_edit.py
```diff
@@ -61,6 +65,37 @@ async def put_flow(
         resp.raise_for_status()
 
 
+async def get_segments(
```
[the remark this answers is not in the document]
samdbmg: I think the simple approach probably helps here, as does the fact it only does a single page (so you can't end up with a _very_ long script run!). I've called that out though.

examples/simple_edit.py
```diff
@@ -152,6 +175,113 @@ async def simple_edit(
...
+        }
+
+        while (len(flow_1_segments) > 0 and len(flow_2_segments) > 0):
+            position_in_flow_timeline = working_time + current_seg["offset"]
+
+            # Draw a segment from the current list (and drop segments if we've passed them)
+            next_seg = current_seg["list"][0]
+            next_seg_tr = TimeRange.from_str(next_seg["timerange"]).normalise(edit_rate.numerator,
+                                                                              edit_rate.denominator)
+            if next_seg_tr.end <= position_in_flow_timeline:
```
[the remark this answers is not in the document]
samdbmg: Yes, and avoids a possible inclusivity bug too

examples/simple_edit.py
samdbmg: Those are all good suggestions: changed

examples/simple_edit.py
samdbmg: I agree about complicating the example, but I have called this out in a comment now

PR #111, opened 2025-02-11
Proposal for an objects endpoint

philipnbbc:
This PR proposes an objects endpoint that includes a list of Flows that references the target media object.

The proposal also adds a property that holds the ID of the Flow that first referenced (via a Flow Segment) the media object. The proposal contains some questions around making the property required rather than optional (as currently specified) and whether a user should be able to set it.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-3550

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1859,19 +1859,16 @@ paths:
           description: Forbidden. You do not have permission to modify this flow. It may be marked read-only.
         "404":
           description: The requested flow does not exist.
-  /objects:
+  /objects/{objectId}:
     head:
       summary: Media Object Information
-      description: |
-        Return flow references and other information about media objects.
-
-        The media object ID is provided by the 'object_id' parameter rather than a path component because the object ID string format is unrestricted.
+      description: Return flow references and other information about media objects.
       operationId: HEAD_objects
       tags:
         - Objects
       parameters:
-        - name: object_id
-          in: query
+        - name: objectId
+          in: path
           description: The media object identifier.
           required: true
           schema:
@@ -1907,8 +1904,6 @@ paths:
       description: |
         Contains flows that references the media object and other information.
 
-        The media object ID is provided by the 'object_id' parameter rather than a path component because the object ID string format is unrestricted.
-
         The paging query parameters and headers are required for the list of flow references in the media object.
         API implementations should return a complete list of flow references within reason and API clients should expect
         paging to happen in some rare cases where a media object is used in many flows.
@@ -1916,8 +1911,8 @@ paths:
       tags:
         - Objects
       parameters:
-        - name: object_id
-          in: query
+        - name: objectId
+          in: path
           description: The media object identifier.
           required: true
           schema:
@@ -1947,7 +1942,7 @@ paths:
               schema:
                 $ref: "schemas/object.json"
         "400":
-          description: Bad request. Invalid query options or missing 'object_id'.
+          description: Bad request. Invalid query options.
         "404":
           description: The requested media object does not exist.
   /flow-delete-requests:
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1862,7 +1862,10 @@ paths:
   /objects:
     head:
       summary: Media Object Information
-      description: Return flow references and other information about media objects.
+      description: |
+        Return flow references and other information about media objects.
+
+        The media object ID is provided by the 'object_id' parameter rather than a path component because the object ID string format is unrestricted.
       operationId: HEAD_objects
       tags:
         - Objects
@@ -1903,6 +1906,8 @@ paths:
       summary: Media Object Information
       description: |
         Contains flows that references the media object and other information.
+
+        The media object ID is provided by the 'object_id' parameter rather than a path component because the object ID string format is unrestricted.
 
         The paging query parameters and headers are required for the list of flow references in the media object.
         API implementations should return a complete list of flow references within reason and API clients should expect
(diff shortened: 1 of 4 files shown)


api/schemas/object.json
```diff
@@ -0,0 +1,28 @@
...
+    "required": [
+        "object_id",
+        "referenced_by"
+    ],
+    "properties": {
+        "object_id": {
+            "description": "The media object identifier.",
+            "type": "string"
+        },
+        "referenced_by": {
```
johnbilt, line 14: In wonder if for understandability the field should include the word flow to make it clearer what the references are.  So something like `flow_references`.  Would also affect line 22 which would become something like `first_flow_reference`

philipnbbc: My decision in the end was to keep it consistent with the existing`collected_by` property name in Flows. However, I agree that adding `flow` would make it more understandable at a glance and there's a chance that objects could be referenced by something else in future. 

Would you be happy with `referenced_by_flows` and `first_referenced_by_flow` instead?

johnbilt: That works for me.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1837,6 +1837,92 @@ paths:
           description: Forbidden. You do not have permission to modify this flow. It may be marked read-only.
         "404":
           description: The requested flow does not exist.
+  /objects:
+    head:
+      summary: Media Object Information
+      description: Return flow references and other information about media objects.
```
j616, line 1843: NIT: Could you add a quick note here on how this uses a query string rather than a path var, and why? I expect it's something that'll catch people out/confuse people fairly frequently. So the more visible we can make it, the better.

docs/adr/0027-add-objects-api-endpoint.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Option 2: Add a `/objects?object_id={object-id}` endpoint that lists all Flows that reference a media object
+
+This `/objects?object_id={object-id}` endpoint provides JSON that lists of all Flows referencing a media object.
+The `referenced_by_flows` property contains the list of Flow IDs.
+
+The Flow Segments that reference the media object can all be found using a GET on `/flows/{flow-id}/segments?object_id={object-id}` for every Flow in the listing rather than every Flow.
+The Flow Segments that reference the media object can all be deleted using a DELETE on `/flows/{flow-id}/segments?object_id={object-id}` for every Flow in the listing.
+
+The `object_id` is as a query parameter rather than a URL path component because the media object ID structure is not restricted.
```
j616, line 50: Should we consider restricting the `object_id` structure so this could be a path parameter? It would be a breaking change. And it would probably have other implications. But it would allow this part of the API to be much more consistent.

j616 (commented):
A couple of questions in line. I'm also wondering if it's worth considering allowing querying by a list of object IDs. Mainly because one of our key use cases is removing all instances of content. This will likely take the form of obtaining a list of object IDs from a flow, querying this endpoint for all those IDs, then calling DELETE against all of those flows with the object IDs. The requests will add up quickly. That said, I don't currently think we expect this feature to be needed often!

docs/adr/0027-add-objects-api-endpoint.md
philipnbbc: The main issue are the slashes as they are expected to also be used to form hierarchies in object storage. Adding them as path components with the slashes would be messy. I guess we could restrict the `object_id` to a UUID string as well for example, but as you say, there may be other implications.

j616:
> Maybe the endpoint should be renamed to `/objects-query`? That is better to allow a `GET /objects-query?object_id=` for a simple single object lookup and `POST /objects-query` for multiple objects. Adding `POST /objects` would be a little confusing / unexpected.
I was thinking a list in the query string. But thinking about it now, I seem to remember that when we recently looked into that for other purposes, we found the options unsatisfactory.

docs/adr/0027-add-objects-api-endpoint.md
j616: Worth still adding it as an option so we've recorded why we rejected it?

philipnbbc:
> > Maybe the endpoint should be renamed to `/objects-query`? That is better to allow a `GET /objects-query?object_id=` for a simple single object lookup and `POST /objects-query` for multiple objects. Adding `POST /objects` would be a little confusing / unexpected.
> > I was thinking a list in the query string.

I didn't think adding potentially hundreds of object IDs into the query string was ideal and that's why a POST seemed better. A POST is also used to make queries to DynamoDB for example.

j616:
> I didn't think adding potentially hundreds of object IDs into the query string was ideal and that's why a POST seemed better. A POST is also used to make queries to DynamoDB for example.
> 


Fair point. Let's leave it for now and see what use cases actually pop up in the real world. I think if we do anything in this space, we should explore how we can provide some level of consistency with the rest of the API.

philipnbbc:
I've changed my mind following James' comments on the query parameter. It is [now](https://github.com/bbc/tams/pull/111/commits/0c493bffb1e2fa2c1da88d8936b6a26bdf5b95c5) a path component of the URL. This matches what the S3 API does for example and also makes it more consistent with the other endpoints.

docs/adr/0027-add-objects-api-endpoint.md
philipnbbc: I've decided to make it a path component - see https://github.com/bbc/tams/pull/111#issuecomment-2704295689

j616 (approved):

PR #115, opened 2025-02-13
appnote: Add appnote covering Authorisation

samdbmg:
# Details
Adds AppNote 0016 proposing approaches to handling authorisation in TAMS, based on the BBC/AWS workshop in Salford in November 2024.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-3534

# Related PRs
Merge after #113 

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,121 @@
...
+| Source Sport A | sport | |
+| Source Sport B | sport | |
+| Source News X  | news, sport_share | This item is shared across to Sport |
+| Source News Y  | news | |
+
+The permissions policies are then implemented as:
+
+- Users in the "sport" user group have read and write access to all content in the "sport" class
+- Users in the "news" user group have read access to all content in the "news" class
+- Users in the "sport" user group have read access to all content in the "news" class
```
j616, line 75: Should this be `sport_share`?

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Implementation
+
+To implement the model above, way to hold the classes in the store is needed, along with a system to store the permissions and how they map on to classes.
+
+For the latter, [Amazon Verified Permissions](https://aws.amazon.com/verified-permissions/) and [Permify](https://github.com/Permify/permify) both serve as permissions management tools.
+They allow authorisation decisions to be made by taking a set of policies defined in some domain-specific language, along with the attributes of the user (group membership) and resource (Source/Flow class), and computing whether to allow the request.
+This decision process is intended to be run inline for each request, for example at an authenticating proxy placed in front of the API server.
+
+For storing classes, an initial proof-of-concept could be built using Source and Flow tags: for example defining "special" tags such as `authz_class.news = 1` (using one tag per class to enable querying for presence of that tag, which is not possible with e.g. a comma-separated list).
```
j616, line 85: Can you add these tags to the known tags listing

j616, line 85: Is the `= 1` here just because the tag needs content?

docs/appnotes/0016-authorisation-in-tams-workflows.md
samdbmg: Yeah, that's a copy-paste error, sorry. Now fixed

docs/appnotes/0016-authorisation-in-tams-workflows.md
samdbmg: Yes: it could equally be `= true`

samdbmg: Good point - [bcdf5f3](https://github.com/bbc/tams/pull/115/commits/bcdf5f39afc6718618eec51d42da1ca8a67cc45c)

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,121 @@
...
```
johnbilt, line 85: I'm concerned that using separate tags for a PoC is not representative of what we're actually do with a single field holding an array.  Need to balance whether we need in a PoC to be able to query content by permission (makes sense longer term with the final API field) v providing something more representative to the permissions layer.

j616: I believe this approach was taken due to limitations of https://aws.amazon.com/verified-permissions/ and https://github.com/Permify/permify , as per line 85. I shall double check this.

j616: I wasn't quite right on the assumptions/limitations. The intention is that authorisation can be implemented as an authorisation proxy in front of the API implementation. The proposal allows that proxy to query the API, particularly the `/flows` or `/sources` endpoints, with a query sting of the form `?tag_exists=authz_class.news` . It is that `?tag_exists` where the limitation lies. The assumption is that this approach would be more efficient and wouldn't require deep integration into the API. I'm going to give this another pass to check those assumptions and decide weather we just need an expanded explanation, a tweak to the app note, or a tweak to the spec.

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,121 @@
...
+6. (Write requests only): Propagate any changes to `authz_class` tags to Flows and Sources collected by this one
+
+## Where to Enforce Authorisation
+
+Some consideration should be given for where to apply the authorisation step, depending on how TAMS is deployed and integrated.
+For example a TAMS instance could be deployed with fine-grained authorisation support, and used directly by systems across an organisation.
+In this case it would make sense to treat all clients of that TAMS instance as identical from an authentication/authorisation perspective: for example a user operating an NLE would be expected to provide suitable authorised credentials, but so too would the organisations MAM when it wants access to the store.
+
+Another deployment approach might see a MAM or other tool expose a TAMS API interface itself, which is proxied through to some simpler backing store.
+In this case the MAM might manage and enforce the other policies and rules around access to content, so it would make more sense to do the same in the TAMS API interface, and then use the MAM's own credentials to access the backing TAMS instance.
```
himslm01: It took me three reads of this sentence to understand it. I think the problem is the 'the' in the section:

... enforce the other policies and rules ...

I think it reads better without 'the':

... enforce other policies and rules ...

docs/appnotes/0016-authorisation-in-tams-workflows.md
j616: This has been updated. The PR now provides adding support for lists as tag values. This allows us to query whether any of a list of values is present in a given tag.

docs/appnotes/0016-authorisation-in-tams-workflows.md
j616: No longer applicable. PR has been updated to add support for lists as tag values, and to use a single tag to store these attributes.

api/TimeAddressableMediaStore.yaml
```diff
@@ -246,9 +290,12 @@ paths:
         - name: tag.{name}
           in: query
           description: |
-            Filter on Sources that have a tag named {name} and with the given value.
+            Filter on Sources that have a tag named {name} with a value in the given comma-seperated list of values.
```
samdbmg, line 413: I guess we don't need to worry about specifying how you give a single item list, because both `?tag.mytag=value` and `?tag.mytag=value,` parse to the same list of one entry?

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
+---
+status: "proposed"
+---
+# Fine-grained Authorisation in TAMS Workflows
+
+## Context and Problem Statement
+
+TAMS provides multiple methods for authentication, as described in [ADR0028](./0028-authentication-methods.md).
+The most commonly used method is currently OAuth2.
```
samdbmg: Nit: This isn't strictly true: we only specify you have to you _Bearer tokens_ for auth, but it doesn't make sense to not use OAuth2 on top of that. I think this applies equally to permissions on URL tokens or basic auth users, but the method of setting scope and mapping permissions is completely out of scope.

How about

```suggestion
The most commonly used method to implement authentication is Bearer tokens acquired using OAuth2 flows.
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+Emerging TAMS use cases are making use of TAMS' media re-use capabilities over increasingly large numbers of users, teams, and organisations.
+As the number of clients accessing content in a TAMS store grows, the need for finer-grained control of that content becomes more acute.
+
+This ADR presents the decisions and considerations that informed the initial approach to fine-grained auth in TAMS.
+
+<!-- This is an optional element. Feel free to remove. -->
+## Decision Drivers
+
+* Be prescriptive enough to enable interoperability of service and client implementations
+* Be permissive enough to facilitate integration with existing workflows and systems
```
samdbmg, line 21: Can we pull out something like this - it's what the second bullet encompasses, but I think it's important to call out this is supposed to work for both ends of that spectrum

> Acknowledge that organisations will have different threat models, and interoperability should be equally possible in more open and more restrictive environments

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Option 1b: Granularity of auth - Flow
+* Option 2b: Level of prescriptiveness - Auth logic
+* Option 3b: Supported architectures - Auth proxy
+* Option 4b: Auth attributes - Single Tag with a list value
+
+This combination of options provides a good balance of well defined behaviour and flexibility.
+It facilitates both interoperability and integration with existing auth systems and workflows.
+The choice of Option 4b allows for us to experiment with, and refine our approach to fine-grained auth.
+Once our approach is mature, we may wish to consider migrating to Option 4c to provide a more efficient solution.
```
samdbmg, line 56: It might be worth noting that we've designed an approach and model for 3b, but it doesn't preclude doing 3a

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+* Good, because it can be implemented with minimal changes to the API specification
+* Good, because it requires the least additional data to be stored in/alongside the API
+* Bad, because it provides limited control over different representations of media
+  * Prevents cost control (e.g. allow access to proxies, but not hi-res)
+  * Prevents management of access to specific storage backends
+
+### Option 1b: Granularity of auth - Flow
+
+Define permissions at the Source and Flow levels.
+Permissions are then propagated down to Segments, and Objects.
```
samdbmg, line 78: Somewhere we should call out that the preferences is by Source, because that matches the rest of the model, but sometimes you might want by-Flow. How about

```suggestion
Permissions are then propagated down to Segments, and Objects.
However the default approach should be to define permissions on Sources (since permissions are likely to apply to all renditions of a piece of content), with the option to use Flows where restricting specific variants is required.
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+  * Time-scoped access requires new flows to be created, potentially via zero-copy mechanism
+
+### Option 1c: Granularity of auth - Segment
+
+Define permissions at the Source, Flow, and Segment levels.
+Permissions are then propagated down to Objects.
+
+* Good, because it can be implemented with minimal changes to the API specification
+* Good, because it provides control over different representations of media
+* Neutral, because it enables direct control of access to segments of flow timelines
```
samdbmg: Should this be a "Good", with a caveat that it is quantised to segments and inter-segment would be harder (e.g. what the sub-bullet says)

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+  * Allowing direct sub-segment access control would require significant modification
+* Bad, because it requires significant amounts of additional data to be stored in/alongside the API
+
+### Option 1d: Granularity of auth - Object
+
+Define permissions at the Source, Flow, Segment, and Object levels.
+
+* Good, because it can be implemented with minimal changes to the API specification
+* Good, because it provides control over different representations of media
```
samdbmg, line 104: Could also include something like:

```suggestion
* Good, because it provides control over different representations of media
* Good, because it maps the right to access content onto the content itself, regardless of how it is re-used
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+Leave the definition of auth logic etc up to individual implementations.
+
+* Good, because it provides high levels of flexibility in implementations
+* Bad, because it makes interoperability around fine-grained auth difficult
+* Bad, because it requires implementers to derive all auth logic from scratch, including a few known non-trivial aspects
+
+### Option 2b: Level of prescriptiveness - Auth logic
+
+Define principles, and high-level auth logic.
+Leave specific algorithms, and the requests permissions evaluation systems may make to the API up to the individual implementations.
```
samdbmg: Nit: Breaking apart that "requests permission evaluation system" with isn't what it said took me a few re-reads 😄 

```suggestion
Leave specific algorithms, and the requests that permissions evaluation systems may make to the API up to the individual implementations.
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Good, because it facilitates interoperability around fine-grained auth
+* Good, because implementers don't have to derive auth logic from scratch
+* Neutral, because it provides medium levels of flexibility to implementations
+
+### Option 2c: Level of prescriptiveness - Specific API requests/pseudocode/scopes
+
+Define all auth logic, algorithms that evaluate that auth logic, and API requests permission evaluation systems will make.
```
samdbmg, line 131: We've gone for recommendations rather than changing the spec (for good reasons): should this 2c talk explicitly about this being "part of the TAMS specification that all implementations must follow" to explore the pros and cons of that distinction?

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Good, because it facilitates interoperability around fine-grained auth
+* Good, because implementers don't have to derive auth logic, or algorithms from scratch
+* Bad, because it provides low levels of flexibility to implementations
+  * This may conflict with existing auth systems, and workflows in deployments
```
samdbmg:
```suggestion
  * This may conflict with existing auth systems, and workflows in deployments and organisational structures and models
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Option 3a: Supported architectures - Deep integration
+
+Assume all TAMS services that support fine-grained auth will implement it with deep integration into the API implementation.
```
samdbmg: 3a/3b: We probably need to expand on "deep integration" vs "auth proxy" - I don't think they're terms of art.

```suggestion
Assume all TAMS services that support fine-grained auth will implement it with deep integration into the API implementation where policy decisions are made by the store implementation as part of processing a request, with full access to any backing databases etc.
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+* Good, because its the most efficient option to implement
+* Bad, because it provides low levels of flexibility when integrating with existing auth systems, and workflows in deployments
+
+### Option 3b: Supported architectures - Auth proxy
+
+Assume fine-grained auth may be implemented using the Auth Proxy pattern.
```
samdbmg: As above
```suggestion
Assume fine-grained auth may be implemented using the Auth Proxy pattern where an HTTP reverse proxy can receive incoming requests, amend them as needed and forward them onto a store (which may have no fine-grained authorisation model). The proxy can use the contents of the original request, the amended request(s) and the response(s) to make a decision on what to return to the user without modifying the underlying store.
```

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+
+Define new parameters throughout the API specification for the purpose of storing auth-specific attributes.
+
+* Good, because it allows for an optimal number API calls when carrying out evaluation of permissions on endpoints
+* Neutral, because it results in increased integration of auth mechanisms into the core specification
+* Bad, because it requires significant changes to the API specification for a feature that has not been widely tested
+
+## More Information
+
+Information on TAMS' approach to authentication may be found in [ADR0028](./0028-authentication-methods.md).
```
samdbmg, line 187: Nit: Maybe this would be better in the context section?

docs/adr/0035-fine-grained-auth.md
```diff
@@ -0,0 +1,173 @@
...
+status: "proposed"
+---
+# Fine-grained Authorisation in TAMS Workflows
+
+## Context and Problem Statement
+
+TAMS provides multiple methods for authentication, as described in [ADR0028](./0028-authentication-methods.md).
+The most commonly used method is currently OAuth2.
+OAuth2 allows for auth tokens to claim [scopes](https://oauth.net/2/scope/) as a means to restrict the permissions of clients and requests.
+This has been used in TAMS implementations to provide coarse grained authorization.
```
samdbmg: Nit: UK spelling (pretty sure I've done this in places!)
```suggestion
This has been used in TAMS implementations to provide coarse grained authorisation.
```

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,298 @@
...
+Users may be assigned combinations of these roles for different purposes, for example:
+
+- `administrator`: Has all four scopes
+- `viewer`: Has `tams-api/read`
+- `editor`: Has `tams-api/read` and `tams-api/write`
+- `store-writer`: Has `tams-api/write`
+- `store-cleanup-system`: Has `tams-api/delete`
+
+To implement the authorisation, the authorisation server checks the requested scopes against the user's access when issuing a token.
+The TAMS server, or it's auth proxy, rejects requests without appropriate scopes.
```
samdbmg: Nit: typo
```suggestion
The TAMS server, or its auth proxy, rejects requests without appropriate scopes.
```

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Finer Grained Authorisation
+
+A further build on the very coarse role-based approach above is to expand the set of permissions to apply to specific Sources and Flows.
+However the implementation of this can become complex and unwieldy, especially if each Source and Flow in the system has a separate set of permissions to manage and it becomes necessary to edit them all to implement a policy change.
+
+Attribute-Based Access Control (ABAC) is one approach to manage this complexity, by describing permissions policies based on the attributes of resources (Sources and Flows), and if necessary, users as well.
+However full ABAC can be challenging to implement and requires a degree of organisational maturity to construct and manage stable attributes.
+This section describes a recommended approach to ABAC authorisation logic to aid interoperability between TAMS implementations.
```
samdbmg: We could clarify exactly how this is and isn't ABAC a smidge here? Also could we weaken down from "recommendation" to open space for something better from the community?

```suggestion
This section describes a possible approach to ABAC authorisation logic to aid interoperability between TAMS implementations, in which content is assigned an attribute in the form of a "class".
```

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,298 @@
...
+| ------------------------------------ | ------------ | --------------------------------------------------------------------------------------- |
+| `/`                                  | `HEAD`/`GET` | Available to all                                                                        |
+| `/service`                           | `HEAD`/`GET` | Available to all                                                                        |
+|                                      | `POST`       | Request must have admin permissions. Otherwise reject.                                  |
+| `/service/storage-backends`          | `HEAD`/`GET` | Available to all                                                                        |
+| `/service/webhooks`                  | `HEAD`/`GET` | Restrict returned data by adding list of claimed auth_classes to `tag.auth_classes.includes`. If the incoming request has `tag.auth_classes.includes` set, the request must be processed with `tag.auth_classes.includes` set to the intersection of the claimed auth_classes and the provided list in `tag.auth_classes.includes`. |
```
samdbmg: I think "includes" is the default behaviour on `tag.auth_classes` in the API change above - so I suspect `.incudes` can be removed throughout?

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,298 @@
...
+
+#### Sources
+
+Read, write, and delete permissions on individual sources may be determined via auth_classes listed in the `auth_classes` tag on the source.
+This may be done via the `/sources/{sourceId}/tags/auth_classes` endpoint.
+
+#### Webhooks
+
+Read, write, and delete permissions on individual webhooks may be determined via auth_classes listed in the `auth_classes` tag on the webhook.
+This may be done via the `/sources/{sourceId}/tags/auth_classes` endpoint.
```
samdbmg: Copy-paste mistake?

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,308 @@
...
+| ------------------------------------ | ------------ | --------------------------------------------------------------------------------------- |
+| `/`                                  | `HEAD`/`GET` | Available to all                                                                        |
+| `/service`                           | `HEAD`/`GET` | Available to all                                                                        |
+|                                      | `POST`       | Request must have admin permissions. Otherwise reject.                                  |
+| `/service/storage-backends`          | `HEAD`/`GET` | Available to all                                                                        |
+| `/service/webhooks`                  | `HEAD`/`GET` | Restrict returned data by adding list of claimed auth classes to `tag.auth_classes.includes`. If the incoming request has `tag.auth_classes.includes` set, the request must be processed with `tag.auth_classes.includes` set to the intersection of the claimed auth classes and the provided list in `tag.auth_classes.includes`. |
```
samdbmg: The implementations has "includes" as the default behaviour - does the `.includes` need to be removed here (and in other similar rows)

samdbmg (commented):
This all looks good overall - few nitpicks and suggested wording tweaks inlined

api/TimeAddressableMediaStore.yaml
j616: This prompted me to check how we handle this in `accept_get_urls`. There we have a regex which prevents the trailing comma. I'll add that regex to these query strings.

docs/adr/0035-fine-grained-auth.md
j616: It is. Though I thought it was also worth sign-posting down here to make it easier to find the link to related documentation.

docs/appnotes/0016-authorisation-in-tams-workflows.md
j616: Woops! Good spot!

docs/adr/0035-fine-grained-auth.md
j616: I feel like that should be a separate, even more prescriptive, option. I'll add something.

j616: Added an option describing this

api/TimeAddressableMediaStore.yaml
```diff
@@ -1962,7 +2036,7 @@ paths:
         "400":
           $ref: '#/components/responses/trait_resource_info_head_400'
         "404":
-          description: The requested media object does not exist.
+          description: The requested media object does not exist. 404 MUST be returned if the ID has been assigned via the storage endpoint, but not yet registered against a segment.
```
danjhd: Not sure if this was deliberate or missing consistency. This extra note has not been added to the 404 response to the `GET` method.

j616: Good spot. Fixed.

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,324 @@
...
+
+- `404` if the request has no permissions on the endpoint
+- `403` if the request has any permission on the endpoint, but not sufficient to complete the request
+
+### Fine-grained authorisation and webhook events
+
+Implementations must evaluate permissions against webhook events themselves as well as the API's HTTP endpoints.
+A basic implementation may enumerate Flows and Sources a user has access to when creating/updating the webhook and use this to filter events.
+This approach is strongly discouraged as permissions may change over time.
+It is recommended that implementations asses permissions on a per-event basis.
```
danjhd: Typo, asses -> assess

docs/appnotes/0016-authorisation-in-tams-workflows.md
```diff
@@ -0,0 +1,324 @@
...
+
+Read, write, and delete permissions on individual sources may be determined via auth classes listed in the `auth_classes` tag on the source.
+This may be done via the `/sources/{sourceId}/tags/auth_classes` endpoint.
+
+#### Objects
+
+Read, write, and delete permissions on individual objects may be determined by filtering returned flows on the object.
+This may be done by setting `flow_tag.auth_classes` to relevant claimed auth classes (e.g. auth classes with read permissions if read permissions on the object are to be verified).
+If `referenced_by_flows` in the returned data is empty, the request DOES NOT have the relevant permissions.
+If `referenced_by_flows` in the returned data is empty, the request DOES have the relevant permissions.
```
danjhd: Should this be "not empty"? At present the condition is the same as the line above so is contradictory.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1937,6 +1999,18 @@ paths:
           required: true
           schema:
             type: string
+        - name: flow_tag.{name}
+          in: query
+          description: |
+            Filter `referenced_by_flows` on tag values. This option is the same as the `tag.{name}` query parameter on the `/flows/` API endpoint.
```
danjhd, line 2210: My gut feeling is that this should also "filter" the flow id from the first_referenced_by_flow field in the same way. Otherwise if we are using this tag query parameter to provide an early implementation of RBAC/ABAC we would be removing a flow id from `referenced_by_flows` only to leave it visible in `first_referenced_by_fMy gut feeling is that this should also "filter" the flow id from the first_referenced_by_flow field in the same way. Otherwise if we are using this tag query parameter to provide an early implementation of RBAC/ABAC we would be removing a flow id from `referenced_by_flows` only to leave it visible in `first_referenced_by_flow`. Just a thought and i'm not "wedded to it".low`. Just a thought and i'm not "wedded to it".

api/TimeAddressableMediaStore.yaml
```diff
@@ -1981,6 +2055,18 @@ paths:
...
```
danjhd, line 2307: As per comment above about also filtering `first_referenced_by_flow`.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1981,6 +2055,18 @@ paths:
...
+          schema:
+            type: string
+        - name: flow_tag_exists.{name}
+          in: query
+          description: |
+            Filter `referenced_by_flows` on tag names. This option is the same as the `tag_exists.{name}` query parameter on the `/flows/` API endpoint.
```
danjhd, line 2313: As per comment above about also filtering `first_referenced_by_flow`.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1937,6 +1999,18 @@ paths:
...
```
danjhd, line 2216: As per comment above about also filtering `first_referenced_by_flow`.

docs/appnotes/0016-authorisation-in-tams-workflows.md
j616: Fixed in https://github.com/bbc/tams/pull/115/commits/1b1ff86db1da6e3c36d931d5dd2627072367b645

api/TimeAddressableMediaStore.yaml
j616: I think you're right. But I think what we choose to do here may be affected by the discussion on whether we are happy to elevate auth classes strait to the core spec. So I'll hold off on this change for now.

PR #117, opened 2025-02-14
johnbilt - Bulk flow segments

johnbilt:
# Details
Ability to add flow segments in bulk via the API.  This affects the existing segments end point due to semantics and existing functionality.  Follows on from ADR 0024

# Jira Issue (if relevant)
Jira URL: N/A

# Related PRs
Related to ADR 0024, option 5.

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1690,7 +1690,7 @@ paths:
     post:
       summary: Create Flow Segment
       description: |
-        Register a new flow segment, attaching the object id given to a point in the flow timeline.
+        Register either a single new flow segment or an array of segments, attaching the object id given to a point in the flow timeline.
 
         The segment may use a newly-written object, or re-use an existing object from another flow.
 
@@ -1706,6 +1706,10 @@ paths:
         If a client needs to modify a Flow segment, e.g. to correct metadata such as the `key_frame_count` or add additional
         URLs to `get_urls`, then the client SHOULD first delete the existing segment and then write a new one. The behaviour is
         undefined if the segment exists and a store may return a 400 error response.
+
+        For successful creation of all segments in the request a 201 response should be provided.  If an error is detected when processing
+        a list of segments then processing should continue to try and process the remaining segments.  A 200 response should be returned 
+        listing the failed segments.
 
         Clients are expected to decide how to break content into media objects, however those objects SHOULD be large
         enough to avoid excessive round trip overheads in the underlying store (_e.g._ of the order of several megabytes)
@@ -1728,11 +1732,23 @@ paths:
             example:
               $ref: examples/flow-segment-post.json
             schema:
-              $ref: schemas/flow-segment.json
+              oneOf:
+              - $ref: schemas/flow-segment.json
+              - type: array
+                items:
+                  $ref: schemas/flow-segment.json
         required: true
       responses:
+        "200":
+          description: Partial success creating segments returning list of failed segments.  
+          content:
+            application/json:
+              schema:
+                $ref: schemas/flow-segment-bulk-failure.json
+              example:
+                $ref: examples/flow-segment-partial-failure.json
         "201":
-          description: created. The flow segment has been created.
+          description: created. The flow segment or list of segments has been created.
         "400":
           description: Bad request. Invalid flow segment JSON or the flow 'container' is not set.
         "403":
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1688,7 +1688,7 @@ paths:
         "404":
           description: The flow ID in the path is invalid.
     post:
-      summary: Create Flow Segment
+      summary: Create Flow Segments
       description: |
         Register either a single new flow segment or an array of segments, attaching the object id given to a point in the flow timeline.
 
--- a/api/examples/flow-segment-partial-failure.json
+++ b/api/examples/flow-segment-partial-failure.json
@@ -5,9 +5,9 @@
             "timerange": "[0:0_10:0)",
             "error": {
                 "type": "TAMSError",
-                "summary": "Flow delete failed because ...",
+                "summary": "Flow segment created failed because ...",
                 "traceback": [
-                  "Exception: The flow segment could not be deleted",
+                  "Exception: The flow segment could not be created",
                   "The above exception was the direct cause of the following exception:",
                   "Traceback (most recent call last):",
                   "  <traceback...>",
--- a/api/examples/flow-segment-partial-failure.json
+++ b/api/examples/flow-segment-partial-failure.json
@@ -1,0 +1,24 @@
+{
+    "failed_segments": [
+        {
+            "object_id": "76f6821b-f85b-4297-af65-dde6a25448a0",
+            "timerange": "[0:0_10:0)",
+            "error": {
+                "type": "TAMSError",
+                "summary": "Flow delete failed because ...",
+                "traceback": [
+                  "Exception: The flow segment could not be deleted",
+                  "The above exception was the direct cause of the following exception:",
+                  "Traceback (most recent call last):",
+                  "  <traceback...>",
+                  "TAMSError: Unable to continue"
+                ],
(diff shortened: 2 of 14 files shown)


philipnbbc (commented):
I would choose option 4 as POSTing a single segment dict rather than an array could be seen to be an API convenience.

If it is an array then the response needs to handle partial failures, issues with the array and there needs to be a description of what a TAMS should do when there are (partial) failures.

Should a server attempt to register all segments or stop at the first failure? First failure might mean a serial registration process which is not optimal. What if there are duplicates in the array? The specifications says the behaviour is undefined if a segment already exists in the store. Should that remain undefined?

The response should probably return all the segments that failed to be registered. This is probably not an API break as the current API doesn't have the segment in the response. Should it be a minimal segment with an error code and description? Should the segments that were successfully registered also be returned?

docs/adr/0029-bulk-flow-segments.md
```diff
@@ -0,0 +1,67 @@
+---
+status: "draft"
+---
+# Ability to bulk create flow segments
+
+## Context and Problem Statement
+
+As part of the work on [ADR0024 source level edit](https://github.com/bbc/tams/blob/main/docs/adr/0024-source-level-edit.md) (also known as edit by reference) it was identified that it could be useful to be able to bulk create flow segments within the API. This ADR explores the impact of [option 5](https://github.com/bbc/tams/blob/main/docs/adr/0024-source-level-edit.md#option-5-provide-an-endpoint-to-bulk-write-flow-segments) this on the API implementation.
```
j616, line 8: Typo?
```suggestion
As part of the work on [ADR0024 source level edit](https://github.com/bbc/tams/blob/main/docs/adr/0024-source-level-edit.md) (also known as edit by reference) it was identified that it could be useful to be able to bulk create flow segments within the API. This ADR explores the impact of [option 5](https://github.com/bbc/tams/blob/main/docs/adr/0024-source-level-edit.md#option-5-provide-an-endpoint-to-bulk-write-flow-segments) on the API implementation. 
```

docs/adr/0029-bulk-flow-segments.md
```diff
@@ -0,0 +1,67 @@
...
+
+Create a new end point, eg `/flows/{flowId}/bulk_segments` to take an array of segments.
+
+* Good - does not break existing implementations
+* Bad - adds additional complexity to the API and implementation
+* Bad - semantics suggests the segments end point should cope with multiple, so could cause confusion between the two end points
+
+### Option 3: Rename existing segments end point and add bulk
+
+Rename the existing `/flows/{flowId}/segments` to be singular `/flows/{flowId}/segment` and then replace the existing `/flows/{flowId}/segments` with the bulk end point
```
j616, line 47: I think another downside with this option is it would potentially create an outlier endpoint without equivalent `HEAD`, `GET`, and `DELETE`. It's also at odds with the rest of the APIs resource endpoints which are plural. Although they are largely of the form `<plaural_resource_type>/<id>`.

docs/adr/0029-bulk-flow-segments.md
```diff
@@ -0,0 +1,67 @@
...
+* Bad - requires all existing ingest solutions to be updated
+* Bad - adds additional complexity to the API and implementation
+
+### Option 4: Allow existing segments end point to accept both
+
+Update the existing `/flows/{flowId}/segments` to accept both the existing dict definition of a single segment as well as an array of segments
+
+* Good - does not break the existing API
+* Good - keeps the number of end points down
+* Bad - adds complexity to the API implementation
```
j616, line 60: I suspect the complexity on the spec side isn't that great. The Spec can handle validation/verification of both variants from a schema point of view (see the flow type "one of" list for the body on the [flows endpoint](https://bbc.github.io/tams/6.0/index.html#/operations/PUT_flows-flowId)). But @philipnbbc has some valid points on the implementation. The alternative would be something like the following pseudo-code:
```
if type(body) is object:
  body = [body]
for segment in body:
  validate_segment(segment)
for segment in body:
  add_segment(segment)
```
It's a little more complexity, but nothing too crazy.

j616:
CI bugs should be sorted in https://github.com/bbc/tams/pull/119 . You may need to re-base for CI to pick up the changes

johnbilt:
> If it is an array then the response needs to handle partial failures, issues with the array and there needs to be a description of what a TAMS should do when there are (partial) failures.
> 
> Should a server attempt to register all segments or stop at the first failure? First failure might mean a serial registration process which is not optimal. What if there are duplicates in the array? The specifications says the behaviour is undefined if a segment already exists in the store. Should that remain undefined?
> 
> The response should probably return all the segments that failed to be registered. This is probably not an API break as the current API doesn't have the segment in the response. Should it be a minimal segment with an error code and description? Should the segments that were successfully registered also be returned?

I'm going to suggest that in the case of the failure of a single segment then the publish should continue to work through the remaining segments.  Unless you're going to make TAMS roll back all the processed segments then you're going to end up with a partial item, so make it as complete as possible.

In terms of response - I think it's a 201 for successfully processing all segments (so works for one or multiple) and then for the partial success then this is more tricky.  Current the best answer is probably a 200 with a list of failed segments in the body.

api/schemas/flow-segment-bulk-failure.json
```diff
@@ -0,0 +1,32 @@
...
+            "description": "The object ID of the segment which has failed to register with the TAMS API",
+            "type": "object",
+            "required": [
+                "object_id"
+            ],
+            "properties": {
+                "object_id": {
+                    "description": "The object ID of the segment which has failed to register with the TAMS API",
+                    "type": "string",
+                    "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
```
philipnbbc, line 19: Remove this pattern as `object_id` is not limited to a UUID string.

docs/adr/0029-bulk-flow-segments.md
```diff
@@ -0,0 +1,75 @@
+---
+status: "draft"
```
philipnbbc, line 2: Remember to change this to "accepted".

api/TimeAddressableMediaStore.yaml
philipnbbc, line 1: Suggestion to change "Create Flow Segment" to "Create Flow Segments" in the /segments POST summary

api/examples/flow-segment-partial-failure.json
```diff
@@ -0,0 +1,24 @@
+{
+    "failed_segments": [
+        {
+            "object_id": "76f6821b-f85b-4297-af65-dde6a25448a0",
+            "timerange": "[0:0_10:0)",
+            "error": {
+                "type": "TAMSError",
+                "summary": "Flow delete failed because ...",
```
philipnbbc, line 8: Nit: here and below, change to something like "flow segment creation failed".

philipnbbc (approved):
A couple comments inline and then LGTM.

api/schemas/flow-segment-bulk-failure.json
```diff
@@ -0,0 +1,32 @@
...
+    "title": "Failed segments",
+    "description": "List of segments that have failed to register",
+    "type": "object",
+    "required": [
+        "failed_segments"
+    ],
+    "properties": {
+        "failed_segments": {
+            "description": "The object ID of the segment which has failed to register with the TAMS API",
+            "type": "object",
```
philipnbbc, line 11: See the actions failure - this should be an array with object items.

philipnbbc: I note that the linter accepted your change but note that in addition to changing the type to `array` it needs an `items` with the `object` type. See for example the [pre property](https://github.com/bbc/tams/blob/45a850c4f379f62903dbe57c18e08ddaaa260180/api/schemas/flow-storage.json#L9)

philipnbbc:
Changes look good, thanks.

PR #122, opened 2025-03-12
ADR for flow image support

johnbilt:
# Details
Adds new flow type to support single still images

# Jira Issue (if relevant)
Jira URL:

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/adr/0031-flow-image-support.md
+++ b/docs/adr/0031-flow-image-support.md
@@ -1,0 +1,54 @@
+---
+status: "draft"
+---
+# Ability to bulk create flow segments
+
+## Context and Problem Statement
+
+Currently the TAMS API has three flow type: Video, Audio and Data.  There are a number of use cases around storing images on the flow timeline to represent where they have been extracted from the content.  These include UI elements, content navigation and VOD workflows.
+
+Currently the nearest option is to treat these as non-continuous video and use the existing flow data model.  This has the issue that the frame rate field is mandatory but does not make sense for still images.  For a regular stream of images (eg 1 per segment) then it could be argued to represent this as 1 over the segment length, however this implies that the image is relevant for the duration of the segment rather than representing a point in time.  Additionally if the images are not evenly spaced (eg shot changes) then this does not make sense.
+
+Separating out the images from the video will also aid discoverability and make it easier for readers to identify the content they want.  It is expected that different processes will be looking for the video (players) and images (ui components) and therefore not be interested in the other media type.  So separating the types will reduce the amount of work required to find the required content.
+
+Most image workflows work with a single image that has been extracted from the video, there are scenarios where a mosaic of thumbnails could be required in a single image.  VOD trick play or AI analysis are examples of this type of workflow.  These have the added challenge of needing to store a manifest which describes the location of each thumbnail and the source timing.  There also may be scenarios where mosaics overlap with the content they contain and therefore break a core TAMS rule that segments should not overlap.
+
+## Considered Options
+
+- Option 1: Add new flow type to support single images
+- Option 2: Add new flow type to support both single and mosaic style images
+- Option 3: Do nothing and continue to put images under a video flow
+
+## Decision Outcome
+
+Recommended to do option 1 at this point and add basic image support in the TAMS API.  Since the mosaic use case is not clearly understood at this point in time then it is recommended that this gets parked until that time.  As it would be additional API fields then option 1 does not prevent this being added in the future.
+
+
+### Implementation
+
+See the API specification changes in PR [#xxx](https://github.com/bbc/tams/pull/xxx).
+
+## Pros and Cons of the Options
+
+### Option 1: Add new flow type to support single images
+
+Add a new flow type to the TAMS specification to support single images only.
+
+- Good: Provides simple method of storing and locating images
+- Good: Correctly described content rather than trying to mis-use existing fields
+
+
+### Option 2: Add new flow type to support both single and mosaic style images
+
+Add the image flow type from Option 1 plus optional metadata fields to describe the content of the mosaic.
+
+- Bad: Full use case not completely clear whether it makes sense to hold mosaics in TAMS or create them downstream from images in the store
+
+
+### Option 3: Do nothing and continue to put images under a video flow
+
+Leave the existing functionality in place and continue to use the existing video flow type and metadata for still images
+
+- Good - No changes required
+- Bad - Mixes content types and makes locating correct flow required harder
+- Bad - Mis-uses metadata fields for frame rate as not required for images but required for a video flow
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,7 @@ For more information on how we use ADRs, see [here](./adr/README.md).
 | [0026](./adr/0026-updated-webhook-events-and-filters.md)           | Updates to the webhook event structures and filters                        |
 | [0029](./adr/0029-bulk-flow-segments.md)                           | Changes to flow segments to add segments in bulk                           |
 | [0030](./adr/0030-allow-external-media-objects.md)                 | Allow a Flow to reference media objects in other Flows and storage         |
+| [0031](./adr/0030-flow-image-support.md)                           | Add new flow type to support single images                                 |
 
 \* Note: ADR 0004a was the unintended result of a number clash in the early development of TAMS which wasn't caught before publication
 


docs/adr/0031-flow-image-support.md
```diff
@@ -0,0 +1,54 @@
+---
+status: "draft"
+---
+# Ability to bulk create flow segments
+
+## Context and Problem Statement
+
+Currently the TAMS API has three flow type: Video, Audio and Data.  There are a number of use cases around storing images on the flow timeline to represent where they have been extracted from the content.  These include UI elements, content navigation and VOD workflows.
+
+Currently the nearest option is to treat these as non-continuous video and use the existing flow data model.  This has the issue that the frame rate field is mandatory but does not make sense for still images.  For a regular stream of images (eg 1 per segment) then it could be argued to represent this as 1 over the segment length, however this implies that the image is relevant for the duration of the segment rather than representing a point in time.  Additionally if the images are not evenly spaced (eg shot changes) then this does not make sense.
```
philipnbbc, line 10: `frame_rate` isn't a mandatory property for video Flows but I kind of agree that the expectation of still images is different from moving images / video.

I imagine that a still image that corresponds to a video frame has a segment timerange matching the associated video frame, e.g. segment timerange is `[50:0_50:20000000)`. A still image can also be instantaneous to support use in compositions with any frame rate or in non-video uses, e.g. segment timerange is `[0:0]`.

docs/adr/0031-flow-image-support.md
```diff
@@ -0,0 +1,54 @@
...
+
+Separating out the images from the video will also aid discoverability and make it easier for readers to identify the content they want.  It is expected that different processes will be looking for the video (players) and images (ui components) and therefore not be interested in the other media type.  So separating the types will reduce the amount of work required to find the required content.
```
philipnbbc, line 12: What should the flow segment timerange recommendation be for a still image that has no timing associated with it? Recommend `[0:0]` or change flow segment `timerange` become optional for POST of image formats and default to `[0:0]` when none is provided? This probably requires an AppNote at some point to describe various cases.

docs/adr/0031-flow-image-support.md
```diff
@@ -0,0 +1,54 @@
...
+- Bad: Full use case not completely clear whether it makes sense to hold mosaics in TAMS or create them downstream from images in the store
+
+
+### Option 3: Do nothing and continue to put images under a video flow
+
+Leave the existing functionality in place and continue to use the existing video flow type and metadata for still images
+
+- Good - No changes required
+- Bad - Mixes content types and makes locating correct flow required harder
+- Bad - Mis-uses metadata fields for frame rate as not required for images but required for a video flow
```
philipnbbc, line 54: `frame_rate` is not required. However, not identifying the content to be still images rather than moving images is probably bad

docs/adr/0031-flow-image-support.md
```diff
@@ -0,0 +1,54 @@
+---
+status: "draft"
+---
+# Ability to bulk create flow segments
```
philipnbbc, line 4: E.g. Supporting image formats

api/schemas/flow-image.json
```diff
@@ -0,0 +1,49 @@
+{
+    "type": "object",
+    "description": "Describes an image Flow",
```
philipnbbc, line 3: Should this specifically say "still" image?

philipnbbc (commented):
Some initial comments

api/schemas/flow-image.json
johnbilt: Done

docs/adr/0031-flow-image-support.md
johnbilt: Updated

api/schemas/flow-image.json
```diff
@@ -0,0 +1,49 @@
...
+            ],
+            "properties": {
+                "format": {
+                    "description": "The primary content type URN for the flow.",
+                    "type": "string",
+                    "enum": [
+                        "urn:x-nmos:format:image"
+                    ]
+                },
+                "essence_parameters": {
```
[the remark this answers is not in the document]
johnbilt: I think matching the essence parameter schemas to the flow types keeps this the most simple as a one to one mapping.  Otherwise I can see challenges about which schema should I use and when.  It also allows for video and images to diverge in the future.  

If there are additional fields we think about be added to the image essence parameters then very happy for people to suggest them however for the initial use cases we had we haven't required any more fields than currently proposed, so were keeping it simple for now.

docs/adr/0031-flow-image-support.md
johnbilt: I think in most cases TAMS will still be paired with a MAM and therefore non time specific files are likely to be stored directly in the MAM with only content which directly relates to a flow timeline being stored in TAMS.  

Using [0:0_0:0] could be an issue when importing file based content into a store which as specific timing data to use.  This often results in content which starts at [0:0_0:0], so could get confused between a still image of the first frame or an image with no timing data.

This is not an issue specific to still images (eg scripts, AQC reports), so going to treat as out of scope of this PR, but probably worth a wider discussion.

docs/adr/0031-flow-image-support.md
johnbilt: My misreading of the spec.  Have removed the comment now.

api/examples/flow-get-200-image-jpeg.json
```diff
@@ -0,0 +1,26 @@
...
+  "created": "2008-05-27T18:51:00Z",
+  "metadata_updated": "2023-09-14T09:45:26Z",
+  "segments_updated": "2023-09-14T09:45:26Z",
+  "description": "Big Buck Bunny",
+  "label": "bbb",
+  "format": "urn:x-nmos:format:image",
+  "created_by": "tams-dev",
+  "updated_by": "tams-dev",
+  "tags": {
+    "input_quality": "full",
```
j616, line 14: `full` isn't one of the values in the [tags listing](https://github.com/bbc/tams/blob/main/docs/appnotes/0003-tag-names.md#input_quality). Does it need adding to that list, or should this value be changed to one of the existing values?

api/schemas/flow-image.json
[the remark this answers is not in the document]
j616: Just looking at the fields for Video, I wonder if it's worth including aspect ratio in the initial set of parameters? Other than that, I think I'm ok with keeping these aspects of the spec separate.

docs/adr/0031-flow-image-support.md
```diff
@@ -0,0 +1,51 @@
...
+## Pros and Cons of the Options
+
+### Option 1: Add new flow type to support single images
+
+Add a new flow type to the TAMS specification to support single images only.
+
+- Good: Provides simple method of storing and locating images
+- Good: Correctly described content rather than trying to mis-use existing fields
+- Good: Provides separation of essence parameters for video and still images so that in the future they can have new fields added independently
+- Bad: TAMS is a basis in the NMOS specification for the flow types.  Since NMOS is for live video feeds then it has no concept of still images.  This ADR proposes using NMOS style terminology to describe the flow type as images, however this is technically not in the NMOS specification.
```
j616, line 37:
```suggestion
- Bad: TAMS uses the NMOS specification as a basis for the flow types.  Since NMOS is for live video feeds then it has no concept of still images.  This ADR proposes using NMOS style terminology to describe the flow type as images, however this is technically not in the NMOS specification.
```

j616, line 37: Is it worth us considering a different namespace for the URN for images? It would be inconsistent across flow types. But it's possibly not a great idea for us to use the NMOS namespace without permission, and for non-NMOS purposes.

j616 (changes requested):

api/schemas/flow-image.json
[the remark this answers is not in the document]
johnbilt: Aspect ratio now added to the flow image schema and to the example

api/examples/flow-get-200-image-jpeg.json
johnbilt: Tags gave been removed from the example as the ones were not relevant.  This came from the video examples which both had this tag and value.  Have updated both to be more accurate.

docs/adr/0031-flow-image-support.md
johnbilt: Updated!

johnbilt: URN for images updated to a tam namespace to remove references to NMOS

api/examples/flow-get-200-image-jpeg.json
j616: Thanks for tidying up these non-standard `input_quality` values. I think there's still a couple of instances of the `full` tag value. https://github.com/bbc/tams/blob/main/api/examples/flows-get-200.json and https://github.com/bbc/tams/blob/main/api/examples/flow-tags-get-200.json .

johnbilt: Also updated now!

j616 (approved):

PR #125, opened 2025-04-17
Specifying storage location when requesting storage allocation

j616:
# Details
This PR adds the ability to to specify storage locations when requesting storage allocation.
This PR unifies references to storage backends at various points in the API.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5406

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1564,6 +1564,7 @@ paths:
             Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             Without `get_urls`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs for example.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: string
             pattern: ^([^,]+(,[^,]+)*)?$
@@ -1575,6 +1576,7 @@ paths:
             An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
             Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             A full list of available `storage_id`s may be found at the `service/storage_backends` endpoint.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: string
             pattern: ^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})(,[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})*$
@@ -1586,6 +1588,7 @@ paths:
             If omitted, both presigned and non-presigned URLs will be returned.
             If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: boolean
         - $ref: '#/components/parameters/trait_resource_paged_key'
@@ -1695,6 +1698,7 @@ paths:
             Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             Without `get_urls`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs for example.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: string
             pattern: ^([^,]+(,[^,]+)*)?$
@@ -1706,6 +1710,7 @@ paths:
             An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
             Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             A full list of available `storage_id`s may be found at the `service/storage_backends` endpoint.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: string
             pattern: ^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})(,[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})*$
@@ -1717,6 +1722,7 @@ paths:
             If omitted, both presigned and non-presigned URLs will be returned.
             If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs.
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: boolean
         - $ref: '#/components/parameters/trait_resource_paged_key'
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -89,7 +89,7 @@ paths:
           $ref: '#/components/responses/trait_resource_listing_head_200'
     get:
       summary: Service Information
-      description: Provide information about the service, including the media store in use.
+      description: Provide information about the service.
       operationId: GET_service
       tags:
         - Service
@@ -121,6 +121,31 @@ paths:
           description: Success. The service info has been updated.
         "400":
           description: Bad request. Invalid service JSON.
+  /service/storage_backends:
+    head:
+      summary: Storage Backend Information
+      description: Return storage backends path headers
+      operationId: HEAD_storage_backends
+      tags:
+        - Service
+      responses:
+        "200":
+          $ref: '#/components/responses/trait_resource_listing_head_200'
+    get:
+      summary: Storage Backend Information
+      description: Provide information about the storage backends available on this service instance. These are populated on deployment of the service instance.
+      operationId: GET_storage_backends
+      tags:
+        - Service
+      responses:
+        "200":
+          description: ""
+          content:
+            application/json:
+              schema:
+                $ref: schemas/storage-backends-list.json
+              example:
+                $ref: examples/storage-backends-get-200.json
   /service/webhooks:
     head:
       summary: List Webhook URLs
@@ -1522,18 +1547,47 @@ paths:
           schema:
             default: false
             type: boolean
+        - name: verbose_storage
+          in: query
+          description: |
+            Include storage metadata in `get_urls`.
+            When `verbose_storage` is `false` only `url`, `presigned`, and `label` will be included in `get_urls`.
+          schema:
+            default: false
+            type: boolean
         - name: accept_get_urls
           in: query
           description: |
             A comma separated list of labels of flow segment `get_urls` to include in the response.
-            Omitting `accept_get_urls` will result in all `get_urls` returned.
+            Omitting `accept_get_urls` will result in no filtering of `get_urls`.
             An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
-            Flow segment `get_urls` with no label cannot be filtered for; they will only be returned if `accept_get_urls` is omitted.
+            Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             Without `get_urls`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs for example.
           schema:
             type: string
             pattern: ^([^,]+(,[^,]+)*)?$
+        - name: accept_storage_ids
+          in: query
+          description: |
+            A comma separated list of `storage_id`s of flow segment `get_urls` to include in the response.
+            Omitting `accept_storage_ids`, or providing an empty `accept_storage_ids` will result in no filtering of `get_urls`.
+            An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
+            Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
+            A full list of available `storage_id`s may be found at the `service/storage_backends` endpoint.
+          schema:
+            type: string
+            pattern: ^([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})(,[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12})*$
+        - name: presigned
+          in: query
+          description: |
+            If set to `true`, only presigned URLs (i.e. those whos `presigned` property is `true`) will be returned in `get_urls`.
+            If set to `false`, only non-presigned URLs (i.e. those whos `presigned` property is `false`) will be returned in `get_urls`.
+            If omitted, both presigned and non-presigned URLs will be returned.
+            If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to
+            generate a large number of pre-signed URLs.
+          schema:
+            type: boolean
         - $ref: '#/components/parameters/trait_resource_paged_key'
         - $ref: '#/components/parameters/trait_paged_limit'
       responses:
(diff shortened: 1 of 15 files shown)


api/schemas/storage-backend.json
```diff
@@ -0,0 +1,34 @@
...
+            "type": "string",
+            "enum": [
+                "http_object_store"
+            ]
+        },
+        "provider": {
+            "description": "The cloud (or other) provider of the storage",
+            "type": "string"
+        },
+        "region": {
```
danjhd, line 17: I am wondering if using "AWS specific" terms here like Region and Availability Zone can be avoided. I don't want the TAMS spec to feel like it only fits with AWS. Do we think there is a way we could replace these fields with some sort of generic "dict" field that would be customisable by the implementation?

j616: When I previously mapped the examples in [AppNote0009](https://github.com/bbc/tams/blob/jamessa-specifyStorage/docs/appnotes/0009-storage-label-format.md#example-mappings), I found that `Region` and `Zone`/`Availability Zone` are used by other cloud providers. e.g. [Google Cloud](https://cloud.google.com/about/locations), [Openstack](https://docs.openstack.org/nova/latest/admin/availability-zones.html), and [Azure](https://learn.microsoft.com/en-us/azure/reliability/availability-zones-overview). For some products and providers (e.g. AWS S3, Azure Blobs, Google Cloud Storage) the storage is region specific but not AZ specific. For others (e.g. AWS S3 Express One Zone), the storage is Region and AZ specific. And for some, particularly on-prem deployments, there may be no region or AZ. I think Region and AZ are universal enough terms, certainly in the Cloud Native domain, that they are useful to include at this level as long as they are optional fields where appropriate.

api/TimeAddressableMediaStore.yaml
johnbilt, line 1: Is there an issue with the ability to filter pre-signed URLs now that it's split from the label to a separate field?  Previously it was part of the label, but now that's not the case I'm not sure if there is a way to request them to be included or excluded from the response.

api/schemas/service.json
johnbilt, line 1: I wonder if it would be worth moving the storage backends to a dedicated endpoint below the service end point.  Thinking ahead to profiles then this would logically fit under the service end point also.  So I think it makes sense for them both to end up as dedicated endpoints.  eg /service/storage_backends and /service/profiles.

Secondly what is the recommended method of populating these end points?  Is it recommended to be done at deployment or do we ever see the need to add/remove these dynamically?  I suspect for storage backends then deployment is probably the better answer, but worth thinking about.

api/TimeAddressableMediaStore.yaml
j616: Yep. The implementation here isn't complete. There should be query params for all of these new parameters where it makes sense.

api/schemas/service.json
j616: I've added a dedicated as suggested. Thinking on when to populate these further, I think it has to be on deployment as the instance will have to be configured with appropriate permissions etc to manage the storage backend.

api/TimeAddressableMediaStore.yaml
j616: Done

danjhd, line 1: Apologies if i have missed this but i did not see any changes to the schema files webhook.json or webhook-post.json. I was expecting to see the new query parameters we have added to the segments endpoint in these webhooks too? Or has a decision been made to save this for a future change?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1522,18 +1547,47 @@ paths:
           schema:
             default: false
             type: boolean
+        - name: verbose_storage
```
danjhd, line 1550: Since this option only modifies the contents of get_urls and HEAD method will not return get_urls should we remove it?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1522,18 +1547,47 @@ paths:
...
+          in: query
+          description: |
+            Include storage metadata in `get_urls`.
+            When `verbose_storage` is `false` only `url`, `presigned`, and `label` will be included in `get_urls`.
+          schema:
+            default: false
+            type: boolean
         - name: accept_get_urls
```
danjhd, line 1558: Since this option only modifies the contents of get_urls and HEAD method will not return get_urls should we remove it?
I realise this is not a change made in the PR but i just noticed this when i commented on the others

api/TimeAddressableMediaStore.yaml
```diff
@@ -1522,18 +1547,47 @@ paths:
...
+            Omitting `accept_get_urls` will result in no filtering of `get_urls`.
             An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
-            Flow segment `get_urls` with no label cannot be filtered for; they will only be returned if `accept_get_urls` is omitted.
+            Flow segment `get_urls` with no label or storage ID cannot be filtered for; they will only be returned if `accept_get_urls` is omitted, and `accept_storage_ids` is omitted or empty.
             Without `get_urls`, the response from the service could be substantially faster if it is not required to
             generate a large number of pre-signed URLs for example.
           schema:
             type: string
             pattern: ^([^,]+(,[^,]+)*)?$
+        - name: accept_storage_ids
```
danjhd, line 1571: Since this option only modifies the contents of get_urls and HEAD method will not return get_urls should we remove it?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1624,18 +1678,47 @@ paths:
...
             type: boolean
+        - name: verbose_storage
+          in: query
+          description: |
+            Include storage metadata in `get_urls`.
+            When `verbose_storage` is `false` only `url`, `presigned`, and `label` will be included in `get_urls`.
+          schema:
+            default: false
+            type: boolean
         - name: accept_get_urls
```
danjhd, line 1691: Is it implied that when all query parameters are supplied they behave as an AND? Just thinking whether it is worthwhile being specific about what happens if accept_get_urls, accept_storage_ids and presigned are all specified by the client at the same time?

api/TimeAddressableMediaStore.yaml
j616: I think we've followed the pattern of HEAD methods always specifying the query strings the GET method accepts. ISTR this is best practice for HEAD requests. If you remember differently, I can do a little research and see if I can confirm one way or the other.

api/TimeAddressableMediaStore.yaml
j616: As above

api/TimeAddressableMediaStore.yaml
j616: As above

api/TimeAddressableMediaStore.yaml
j616: Yes. I think that would be the expected behaviour. I'll add clarification.

api/TimeAddressableMediaStore.yaml
j616: Good spot. It had slipped my mind. I'll add that in.

api/TimeAddressableMediaStore.yaml
danjhd: Fair enough it actually makes the implementation easier if the query parameters are the same, it was just a thought that occurred to me when reviewing it. I am happy to leave them in for that reason.

api/TimeAddressableMediaStore.yaml
danjhd: As above too

api/TimeAddressableMediaStore.yaml
danjhd: As above too

docs/adr/0032-specifying-storage-backend.md
```diff
@@ -0,0 +1,236 @@
...
+* Unify references to storage types across the spec
+  * Storage backends are currently referenced in the [AppNote0009](https://github.com/bbc/tams/blob/main/docs/appnotes/0009-storage-label-format.md) label structure, and the `media_store` parameter at the `/service` endpoint in different ways
+  * This ADR would result in the addition of further references when selecting storage backends when allocation storage
+* Fulfil the elevation of the components of the [AppNote0009](https://github.com/bbc/tams/blob/main/docs/appnotes/0009-storage-label-format.md) label structure to the core API as planned in [ADR0021](https://github.com/bbc/tams/blob/main/docs/adr/0021-storage-label-format.md)
+* Avoid further potential breaking changes to references to storage backends if the elevation of the components of the [AppNote0009](https://github.com/bbc/tams/blob/main/docs/appnotes/0009-storage-label-format.md) label structure to the core API happened at a later date
+
+## Considered Options
+
+* Option 1a: Multiple separate TAMS instances using the existing API with no changes
+* Option 1b: Update the TAMS API to allow storage backend to be specified when requestion storage allocation
```
GeorginaShippey: NIT: requestion -> requesting

docs/adr/0032-specifying-storage-backend.md
```diff
@@ -0,0 +1,236 @@
...
+* Option 4b: Elevate the components of [AppNote0009](https://github.com/bbc/tams/blob/main/docs/appnotes/0009-storage-label-format.md) label structure to core spec
+* Option 5a: Don't add any additional filters on the flow `/segments`
+* Option 5b: Add a storage backend ID filter to the flow `/segments` endpoint
+* Option 5c: Add a pre-signed filter to the flow `/segments` endpoint
+
+## Decision Outcome
+
+Chosen options:
+
+* Option 1b: Update the TAMS API to allow storage backend to be specified when requestion storage allocation
```
GeorginaShippey: NIT: requestion -> requesting

docs/adr/0032-specifying-storage-backend.md
```diff
@@ -0,0 +1,236 @@
...
+This option may also imped the discovery and re-use of content across TAMS instances.
+
+* Good, because it doesn't require a spec change
+* Neutral, because clients may have to be able to communicate with multiple TAMS instances anyway when crossing business boundaries
+* Bad, because resource use is poor
+* Bad, because it relies on off-spec management of the lifecycle of Objects in some cases
+* Bad, because it requires clients to communicate with multiple TAMS instances for more use cases
+* Bad, because it impedes zero-copy re-use of Segments
+
+### Option 1b: Update the TAMS API to allow storage backend to be specified when requestion storage allocation
```
GeorginaShippey: NIT: again requestion -> requesting, assuming this has been copied and pasted around. But it's making me requestion myself!

j616: Well thats silly.... fixed

api/examples/storage-backends-get-200.json
johnbilt, line 1: At the risk of looking too far into the future - do we think that all store types will have the same set of fields, eg region, az etc.  Would there be a benefit of keeping the id/label/store_type at the top level, but moving the fields which are specific to the type to a second level so they could be changed in the future?

Initial use case I could see would be a remote TAMS store where the store_type would be different, but then you'd probably want something like URL of the remote store and maybe other details.

j616: Something like this?
```json
{
    "id": "60af2ab4-e8a5-4c65-a09b-d35983680315",
    "label": "example-store-name",
    "default_storage": true,
    "store_type": "http_object_store",
    "store_properties": {
        "provider": "example-cloud-provider",
        "region": "eu-west-1",
        "availability_zone": "a",
        "store_product": "example-storage-product"
    }
}
```
Its an interesting thought. These params are all optional, though. And many may be applicable to other store types. So we'd have to try and preempt what would be common across store types, and what would be unique to `http_object_store`.

I'll have a think on it.

johnbilt: Something like that was what I was wondering.  Means they could then have different properties for different storage types without ending up with the possibility of people declaring weird permutations of future fields

api/TimeAddressableMediaStore.yaml
```diff
@@ -121,6 +121,31 @@ paths:
...
+      description: Return storage backends path headers
+      operationId: HEAD_storage_backends
+      tags:
+        - Service
+      responses:
+        "200":
+          $ref: '#/components/responses/trait_resource_listing_head_200'
+    get:
+      summary: Storage Backend Information
+      description: Provide information about the storage backends available on this service instance. These are populated on deployment of the service instance.
```
GeorginaShippey, line 136: Do you think in the future we will allow creation/deletion of storage backends through the API? I think deploy time is a good choice for now, but I'm feeling that the storage backends wont be completely static over the lifetime of a TAMS instance.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1522,18 +1547,50 @@ paths:
...
+            Where multiple filter query parameters are provided, the returned `get_urls` will match all filters.
           schema:
             type: string
             pattern: ^([^,]+(,[^,]+)*)?$
+        - name: accept_storage_ids
+          in: query
+          description: |
+            A comma separated list of `storage_id`s of flow segment `get_urls` to include in the response.
+            Omitting `accept_storage_ids`, or providing an empty `accept_storage_ids` will result in no filtering of `get_urls`.
+            An empty `accept_get_urls` results in an empty or no `get_urls` in the response.
```
GeorginaShippey: Should this line be here? Looks like a copy and paste artefact about accept_get_urls

api/TimeAddressableMediaStore.yaml
```diff
@@ -1624,18 +1681,50 @@ paths:
...
```
GeorginaShippey: Again here. Should this line be here? Looks like a copy and paste artefact about accept_get_urls

docs/adr/0032-specifying-storage-backend.md
```diff
@@ -0,0 +1,236 @@
...
+---
+# Specifying storage backend when requesting storage allocation
+
+## Context and Problem Statement
+
+The TAMS API supports the referencing of multiple storage backends for Flow Segments in the [`get_urls`](https://bbc.github.io/tams/main/index.html#/operations/GET_flows-flowId-segments#response-body) list.
+Properties of these locations may be signalled in the label as per [Application Note 0009](https://github.com/bbc/tams/blob/main/docs/appnotes/0009-storage-label-format.md).
+The support for multiple locations in `get_urls` was initially created to allow reference to locations outside of a given TAMS deployment (e.g. in another TAMS deployment, or another system entirely).
+The [allocation of storage](https://bbc.github.io/tams/main/index.html#/operations/POST_flows-flowId-storage) by a TAMS instance does not currently make any provisions for multiple storage backends.
+It currently is up to the TAMS implementation how and where to allocate storage.
```
GeorginaShippey: NIT: Missing 'to decide'
It currently is up to the TAMS implementation **_to decide_** how and where to allocate storage.

api/schemas/storage-backends-list.json
```diff
@@ -0,0 +1,29 @@
...
+          "store_product",
+          "id"
+        ]
+      },
+      {
+        "type": "object",
+        "properties": {
+          "default_storage": {
+            "description": "If set to `true`, this is the default storage backend. The default storage backend will be used if the client does not specify a storage backend id when requesting the allocation of storage. If this parameter is not set, assume `false`.",
+            "type": "boolean"
```
GeorginaShippey, line 31: I can't find but may have missed any guidance where it is mentioned that a TAMS implementation should have only one default storage_backend - is that true and should we document that anywhere?

docs/adr/0021-storage-label-format.md
```diff
@@ -1,6 +1,7 @@
 ---
-status: "accepted"
+status: "superseded by ADR-0032"
```
GeorginaShippey, line 2: I'm beginning to think we should be indicating this sort of thing in the table of Supporting Documentation. This could be something we think about when we are brushing up the docs.

api/schemas/storage-backend.json
```diff
@@ -0,0 +1,39 @@
...
+    "description": "Provides technical, and logic metadata about a storage backend",
+    "title": "Storage Backend",
+    "properties": {
+        "id": {
+            "description": "Storage backend identifier",
+            "type": "string",
+            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
+        },
+        "label": {
+            "description": "Freeform string label for a storage backend. Will be be included alongside URLs referencing the storage backend in segment `get_urls`.",
```
GeorginaShippey: NIT: double be.
Will be be included -> Will be included

I also a have a little trouble with this one. 'Will be included alongside URLs referencing the storage backend in segment 'get_urls'' - isn't all the information in this .json included alongside the URLs referencing the storage backend in the segment get_urls? I'm just trying to understand why this gets a special shout out / what verbose_storage should actually show.

docs/appnotes/0009-storage-label-format.md
```diff
@@ -1,4 +1,7 @@
-# Storage label format specification
+# 0009: Storage label format specification
+
+> [!CAUTION]
+> DEPRECATED: See [ADR0032](https://github.com/bbc/tams/blob/main/docs/appnotes/0032-specifying-storage-backend.md) for more details
```
GeorginaShippey, line 4: Again, when we look at brushing up docs, it'd be good to reflect this in the table of supporting documentation.

api/TimeAddressableMediaStore.yaml
j616: As per https://github.com/bbc/tams/pull/125#pullrequestreview-2896837044 . I don't think so. A client instance will have to be provisioned with appropriate credentials etc to manage the storage backend. As such, I think this has to be a deploy-time configuration and not run-time.

api/TimeAddressableMediaStore.yaml
j616: This was a copy-paste error. Shall remove.

api/TimeAddressableMediaStore.yaml
j616: As above.

docs/adr/0032-specifying-storage-backend.md
j616: Shall fix

api/schemas/storage-backends-list.json
j616: I previously had something to this effect here. I removed it but can't find any notes as to why! I think you're right though. I'll add it back.

docs/adr/0021-storage-label-format.md
j616: Good idea

api/schemas/storage-backend.json
j616: A quirk of the existing spec is that there's no clear way that we allow clients to set labels for controlled URLs, but we do allow them to be set for uncontrolled URLs and returned against any URL. We say that attempting to update segments in-place, including the get_urls list and properties, can result in undefined behaviour and may return a 400 error. The offshoot of this is that implementations would normally set the label on controlled URLs themselves according to AppNote 0009. And AppNote 0009 was only ever advisory, meaning that implementations could, in theory, set a label that wasn't tied to the storage backend at all. 

This is here to make this more consistent and explicitely call out that the label is consistent for a given storage backend for controlled URLs, without treading on the fact that labels can be set by clients for uncontrolled URLs. That said, it looks like I'm unintentionally overriding the `label` description in the flow segments schema. I think that's probably unintentionally useful. I'll modify that to call out the slightly different behaviour for controlled and uncontrolled URLs.

j616: I'll remove this sentence from here and include a description of the intended behaviour in the description of the label at the segments endpoint.

docs/appnotes/0009-storage-label-format.md
j616: Done

api/TimeAddressableMediaStore.yaml
GeorginaShippey: We had a short discussion in private on how adding new backends could be implemented in practice at runtime, but came to the conclusion that a lack of ability to reconfigure backends at runtime would encourage implementers to use redeployments/rolling upgrades avoiding long running static deployments. YAGNI until enough people need it!

j616: I've added discussion of these options to the ADR

api/TimeAddressableMediaStore.yaml
```diff
@@ -121,6 +121,31 @@ paths:
           description: Success. The service info has been updated.
         "400":
           description: Bad request. Invalid service JSON.
+  /service/storage_backends:
```
danjhd: Perhaps a nit here but perhaps we should use storage-backends (hyphen in place of underscore) to stay consistent with the fact we have used hyphen already? (flow-delete-requests)

j616: Good point. We also have `read_only` but that is derived from a parameter. :upside_down_face:

j616: I've changed it to `storage-backends`

PR #130, opened 2025-05-20
Flow profiles

### LINKED ISSUE #190: Flow Profiles
**Description**
Provide a way to specify the shape/technical characteristics of a type of Flow and assign it an identifier and a name as shorthand.

That could be useful for finding Flows that can be cut together without transcoding in a lightweight edit.

It could also drive UI elements in ingest/transcode/etc tools to setup encoders

**Use Case**
As a tool that implements lightweight edit, I want to identify and select suitable Flows that can be cut together, because they have matching characteristics.
As a systems integrator, I want to signal to my users what the "house styles" are in this deployment.

johnbilt:
# Details
Add the content of flow profiles to aid with the creation of flows in stores where there are house formats being used.  Also significantly simplifies matching flows across edit by reference workflows

# Jira Issue (if relevant)
Jira URL:

# Related PRs
If merged before https://github.com/bbc/tams/pull/115, add the new endpoints to the auth logic listings in that PR. If merged after, add that logic in this PR

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+
+The use of house formats then makes workflows such as ingest, rendering new content after editing, generation of proxy content and edit by reference easier it defines one or more common formats for all the systems to utilise.
+
+### Edit by reference flow matching
+
+For edit by reference workflows there is a need both to create new sources and flows for the new item and also combine the segments together to create the new edit.
+
+If the source content is only coming from a one source (potentially with multiple flows), for example a simple clip, then it is relatively easy to clone the technical parameters of the existing flows and then join the segments to the new flow.
+However if the content for the new edit is derived from multiple sources then there is a need to match the flows from the different sources based on the technical characteristics.
+Currently the TAMS API does provide an easy method to do this, so it is necessary to compare all the flows in code to find the matches.
```
himslm01: > Currently the TAMS API does provide

Do you mean .... does **not** ... ???

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Challenges
+
+### House formats
+
+Within a customer implementation of TAMS it is expected that there would be limited number of recommended formats for users to work in. 
+Typically this could include a target format for the high quality media and standardised formats for proxy media and images.
+These formats are typically refered to as House formats and do no preclude the storing of other content within the store, however it is likely to be normalised at some point to the house format.
+
+The use of house formats then makes workflows such as ingest, rendering new content after editing, generation of proxy content and edit by reference easier it defines one or more common formats for all the systems to utilise.
```
himslm01: > edit by reference easier it defines one or more

I think ... *edit by reference easier. It defines one or more* ...
Or perhaps ... *edit by reference easier. House formats define one or more* ...

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
+# 00xx: DRAFT: Using Flow Profiles
+
+## Abstract
+
+The TAMS API is deliberately agnostic to the format of the media which it is managing to allow for future compatibility with new formats without needing to change the API specification.
```
himslm01: > which it is managing to allow

Perhaps some punctuation, please.

How about: ... *which it is managing, to allow* ...

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+If the source content is only coming from a one source (potentially with multiple flows), for example a simple clip, then it is relatively easy to clone the technical parameters of the existing flows and then join the segments to the new flow.
+However if the content for the new edit is derived from multiple sources then there is a need to match the flows from the different sources based on the technical characteristics.
+Currently the TAMS API does provide an easy method to do this, so it is necessary to compare all the flows in code to find the matches.
+The more sources involved in this process, the harder this becomes.
+
+## TAMS Profiles within the API
+
+The TAMS profile model is split into a number of stages:
+
+* The list profiles end point which describes the recommended profiles supported by the store
```
himslm01: > The list profiles end point which

Seems to be making assumptions on the name of the end point before introducing it.

Maybe: *An end-point which...*

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+* The ability to supply a profile when creating a flow and the de-normalisation of the technical details at the point of creation
+* The ability to query flows for a given profile to easily and quickly find the required media type
+
+### Profiles endpoint
+
+It is possible to create, view, edit and delete profiles via the dedicated API calls under the TAMS service end point (/services/profiles).
+These profiles hold all the technical parameters required to create a flow in a single location.
+For a system looking to ingest standardised content into TAMS then it would expect to create content matching one or more profiles as defined in the endpoint.
```
himslm01: > For a system looking to ingest standardised content into TAMS then it would

This sentence seems clumsy.

Maybe: *A system looking to ingest standardised content into TAMS would* ...

himslm01 (commented):
A few readability improvements.

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+The TAMS API is deliberately agnostic to the format of the media which it is managing to allow for future compatibility with new formats without needing to change the API specification. 
+This provides challenges for interoperability and workflows such as edit by reference where standard media formats are required across multiple sources.
+
+This document describes how flow profiles can be used within the TAMS API to simplify workflows while maintaining the flexibility of the core API.
+
+## Challenges
+
+### House formats
+
+Within a customer implementation of TAMS it is expected that there would be limited number of recommended formats for users to work in.
```
j616:
```suggestion
Within a customer deployment of TAMS it is expected that there would be limited number of recommended formats for users to work in. 
```
I think this is what you mean? Rather than an implementation of a specific store/client.

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+When using the second option on submission of the create flow request, the store will then read the metadata for a given profile and use this to populate the metadata for the flow.
+This de-normalisation of the technical parameters means that on the read side the flows created via both mechanisms have the same technical parameters available.
+Additionally if the decision is made to change a standard profile in the future then this does not affect the existing flows as the parameters have already been replicated and should remain unchanged to reflect the media actually being stored.
+
+Flows that have been created from a profile will include the parameter indicating which profile they were created from.
+This differentiates them from the flows that have been created with the technical characteristics directly.
+
+### Query flows using a profile
+
+The get flows endpoint has the ability to query for flows with a given profile.  When combined with other query parameters such as source id then this means it is easy to start matching content formats.
```
j616: I know its a little more work on the API side, but should querying by profile match the parameters rather than the profile name? That way you could match flows which just so happen to meet the requirements for a specific profile. And if a profile is re-defined such that it breaks compatible (non-ideal, but possible under this design) then you wouldn't run the risk of getting non-compatible flows back.

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+To aid interoperabity a set of recommended profiles is hosted as part of the API specification Github respository.
+These are profiles which have been tested and used in real workflows and are recommended as a starting point for implementors building a new store.
+It is recommended that tools which write to the TAMS API try and support as many of the recommended profiles of their target content types as possible.
+
+Additional recommened profiles will be added to the repository as new use cases are identified or new media formats added.
+
+## Suggested metadata fields in the profile end point
+
+- Identifier (format to be decided)
+- Type (video/audio/image/data)
```
j616: Should this include multi? More broadly, would it be useful to be able to define a set of flows which combined match a profile? That might be too complicated, though.

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,94 @@
...
+- Label
+- Description
+- Version (recommended to increment when changing)
+- Created/Created by
+- Modified/Modified by
+- Tags
+- Technical characteristics (first level flow properties such as codec, wrapper, target chunk length etc, plus essence parameters)
```
j616: I suspect the ideal would be for us to re-use the params definition from the flows so they're not defined in two places. This might need a minor re-factor, but that shouldn't be too difficult.

johnbilt: That's exactly my plan - the technical characteristics should be a 1:1 mapping with the flow model so that the API internally can basically just take that and apply it to the flow that's being created.  

I do have some examples I need to add to the PR which hopefully will make this clearer

docs/appnotes/00xx-using-flow-profiles.md
johnbilt: The multi as a profile in it's own right doesn't make a lot of sense as there are basically no parameters that you're bringing across from the profile into the flow.  

The multi as a set of profiles to be created is an interesting idea, however I think it would be more tricky to manage.  For example, what happens if you want to re-use the same audio profile across different multi combinations?  One to ponder on!

docs/appnotes/00xx-using-flow-profiles.md
johnbilt: One of my open questions is the best way to link the profile and the flow - is it by name or by ID.  As you mention, what happens if the profile is updated - should that actually be a new profile since it's different.  I have version currently in my draft profile model, but this adds more complexity to match two fields.

Definitely one for further discussion and then update the ADR with the outcome

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,95 @@
...
+
+This document describes how flow profiles can be used within the TAMS API to simplify workflows while maintaining the flexibility of the core API.
+
+## Challenges
+
+### House formats
+
+Within a customer deployment of TAMS it is expected that there would be limited number of recommended formats for users to work in. 
+Typically this could include a target format for the high quality media and standardised formats for proxy media and images.
+These formats are typically refered to as House formats and do no preclude the storing of other content within the store, however it is likely to be normalised at some point to the house format.
```
samdbmg: Nit: typo
```suggestion
These formats are typically refered to as House formats and do not preclude the storing of other content within the store, however it is likely to be normalised at some point to the house format.
```

docs/appnotes/00xx-using-flow-profiles.md
```diff
@@ -0,0 +1,95 @@
...
+
+The use of house formats then makes workflows such as ingest, rendering new content after editing, generation of proxy content and edit by reference easier.
+House formats define one or more common formats for all the systems to utilise.
+
+### Edit by reference flow matching
+
+For edit by reference workflows there is a need both to create new sources and flows for the new item and also combine the segments together to create the new edit.
+
+If the source content is only coming from a one source (potentially with multiple flows), for example a simple clip, then it is relatively easy to clone the technical parameters of the existing flows and then join the segments to the new flow.
```
samdbmg: I think we might need to find a word that isn't "source" here, because it collides with the idea of a Source as a set of Flows of the same thing. Maybe "input" is better?

docs/appnotes/00xx-using-flow-profiles.md
samdbmg: Back in November we talked about the idea of `presets`: a set of profiles you create together for some defined task. I think where we got to is that should be out of scope in TAMS: you'd define it as part of some other system in terms of profiles, based on your standard process for ingest/proxy transcode etc. (and if necessary, that system would create Flows for you and just hand you some IDs to write to, with appropriately scoped credentials)

docs/appnotes/00xx-using-flow-profiles.md
samdbmg: Of course, the "standard" way to do this is to make profiles immutable (possibly with a version if you do want to mutate)? That said I do like the idea of the "greedy match" described above where even if you don't refer to a profile, if you create a Flow that matches it, you can pick that Flow up in profile queries.
I suppose that could be something the store itself implements though: even if you don't supply a `profile` at Flow creation, if it matches then the store fills it in

api/schemas/profile.json
```diff
@@ -0,0 +1,53 @@
...
+    "id",
+    "flow_metadata"
+  ],
+  "properties": {
+    "id": {
+      "description": "Profile identifier.",
+      "type": "string",
+      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
+    },
+    "urn": {
```
samdbmg: Do we have too many IDs here? Style thus far is to have a bare UUID (this is annoying in it's way, because you can't tell what the ID is for, but changing that would be rather disruptive!) and then use that UUID or the URL that contains it where you need a strong reference, with the label providing human-readability.
URN here kinda blurs the line between the two: should we try and collapse the number of ID's in use somehow, or change the primary key of a profile?
(There's also the small matter that [RFC 6648](https://datatracker.ietf.org/doc/html/rfc6648) deprecates the `x-` prefix for this, but there's precedent in the style used for `format` in Flows)

api/schemas/profile.json
```diff
@@ -0,0 +1,53 @@
...
+
+    "flow_metadata": {
+      "description": "The technical characteristics of the profile.  This section will be mapped directly to all flows created using this profile.",
+      "type": "object",
+      "oneOf": [
+        { "$ref": "flow-video.json" },
+        { "$ref": "flow-audio.json" },
+        { "$ref": "flow-image.json" },
+        { "$ref": "flow-data.json" },
+        { "$ref": "flow-multi.json" }
```
samdbmg: Do we want to be able to create a multi profile? https://github.com/bbc/tams/pull/130/files#r2126897787 is inconclusive on this

api/examples/profile-post-201.json
```diff
@@ -0,0 +1,37 @@
+{
```
samdbmg, line 1: Nit: This and `api/examples/profile-get-200-video.json` are identical - I wonder if they can be collapsed to a single file serving both purposes?

api/examples/profile-post-201.json
```diff
@@ -0,0 +1,37 @@
...
+  "id": "329b98d6-adeb-418c-8eaf-eca14edc1706",
+  "urn": "urn:x-tam:profile:video_50m_1080p50_ts",
+  "label": "1080p50 TS Video 50Mpbs",
+  "description": "50Mbps HD video",
+  "created_by": "tams-dev",
+  "created": "2008-05-27T18:51:00Z",
+  "tags": {
+    "_ffmpeg_command": "...."
```
samdbmg: 💯 I really like the idea of including this - should the suggestion be mentioned in the appnote/documentation somewhere?

api/schemas/flow-put.json
```diff
@@ -0,0 +1,37 @@
...
+            { "$ref": "flow-audio.json" },
+            { "$ref": "flow-image.json" },
+            { "$ref": "flow-data.json" },
+            { "$ref": "flow-multi.json" }
+          ]
+        }
+      ]
+    }
+  ],
+
```
samdbmg: Nit: typo
```suggestion
  ]
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,200 @@ paths:
...
+      description: Return profile backends path headers
+      operationId: HEAD_profiles
+      tags:
+        - Profiles
+      parameters:
+        - name: profile_id
+          in: query
+          description: Filter on source identifier.
+          schema:
+            $ref: '#/components/schemas/uuid'
```
samdbmg: Given there's a `/service/profiles/<profile_id>` endpoint (and it's a primary key) maybe including this as well doesn't make as much sense, and there should just be one way to do that?

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,200 @@ paths:
...
+      description: List of profiles supported by the store instance for use when creating flows.
+      operationId: GET_profiles
+      tags:
+        - Profiles
+      parameters:
+        - name: profile_id
+          in: query
+          description: Filter on source identifier.
+          schema:
+            $ref: '#/components/schemas/uuid'
```
samdbmg: As above

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,200 @@ paths:
...
+            schema:
+              $ref: schemas/profile.json
+        required: true
+      responses:
+        "201":
+          description: The flow has been created.
+          content:
+            application/json:
+              example:
+                $ref: examples/profile-put-200.json
```
samdbmg: Nit: This example might need a rename (it's a POST now, and it's a 201)

api/TimeAddressableMediaStore.yaml
```diff
@@ -924,7 +1128,17 @@ paths:
           description: Bad request. Invalid query options.
     put:
       summary: Create or Replace Flow
-      description: Create or replace the flow metadata.
+      description: |
+        Create or replace the flow metadata.
+
+        Flows can be created either by directly supplying all the required technical metadata or by referencing
+        an existing profile in the store.  When using a profile the store will combine the supplied flow metadata
+        with the profile metadata such that on reading the flow a complete set of flow metadata is supplied.  It
```
samdbmg: "When using a profile" can unintentionally be read as a way to override the profile settings (because it's not clear that "flow metadata" and "profile metadata" don't overlap). The restriction on doing so is explained in the second paragraph, but does this better convey that message here? Or maybe the notion that trying to modify a profile at Flow creation is an error should be moved up a bit?

> When using a profile, the store will use the technical details (`format`, `codec`, `essence_parameters` etc.) from that profile, along with the metadata provided.

samdbmg (commented):
Looks good to me overall: a good approach of maintaining flexibility while constraining in a way to make it easier to use.

I've inlined a handful of suggestions, and a couple of nits

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,200 @@ paths:
...
+                video:
+                  summary: Video Profile
+                  value:
+                    $ref: examples/profile-get-200-video.json
+                jpegimage:
+                  summary: Image Profile
+                  value:
+                    $ref: examples/profile-get-200-image.json
+        "404":
+          description: The requested flow does not exist.
```
GeorginaShippey:
```suggestion
          description: The requested profile does not exist.
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,200 @@ paths:
...
+          content:
+            application/json:
+              example:
+                $ref: examples/profile-put-200.json
+              schema:
+                $ref: schemas/profile.json
+        "400":
+          description: Bad request. Invalid profile JSON or profile already exists.
+        "404":
+          description: The requested Flow ID in the path is invalid.
```
GeorginaShippey:
```suggestion
          description: The requested profile ID in the path is invalid.
```

api/schemas/flow-put.json
```diff
@@ -0,0 +1,37 @@
...
+  "title": "Flow",
+  "allOf": [
+    {
+        "$ref": "flow-common.json"
+    },
+    {
+      "oneOf": [
+        {
+          "type": "object",
+          "required": [
```
GeorginaShippey: Adding a "title" here would make things a little easier to read on the rendered version e.g. "Profile Flow". I also think the two 'OneOf's can be collapsed in to one containing the profile and all the other flow types.

api/schemas/flow-technical.json
```diff
@@ -0,0 +1,53 @@
+{
+  "type": "object",
+  "description": "Describes the technical characteristics of a Flow (common properties to all Flows, imported by type-specific specifications)",
+  "title": "Flow Core",
```
GeorginaShippey: NIT: Would be good to have the title updated as it keeps the rendered version more readable, there are currently two schemas rendered out as Flow Core

api/schemas/flow-common.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "object",
-  "description": "Describes a Flow (common properties to all Flows, imported by type-specific specifications)",
+  "description": "Describes the core fields of a Flow (common properties to all Flows, imported by flow-get and flow-put specifications)",
   "title": "Flow Core",
```
GeorginaShippey: NIT: Would be good to have the title updated as it keeps the rendered version more readable, there are currently two schemas rendered out as Flow Core

docs/appnotes/00xx-using-flow-profiles.md
johnbilt: Let's keep the scope of this PR to just profiles to get it into the spec.  If we want to create presets to hold multiple profiles then we handle that as a follow on PR at that point.

api/schemas/profile.json
johnbilt: The reason for the URN is so that we can create profiles which work across multiple stores (federation/replication models).  So the ID becomes an internal store ID, where the URN is a wider identifier.  Part of the idea was then to ultimately host a set of core profiles on the github that people could then refer to.

Happy to have a chat about this further as I do agree it's a little messy having two identifiers, but not sure of a better model at the moment

api/schemas/profile.json
johnbilt: Good catch - a profile for a multi makes no sense.  Even allowing for pre-sets I'm not sure that you'd necessarily use the existing multi schema for that

api/examples/profile-post-201.json
johnbilt: It's already in the app note.  Added a new note to the PUT endpoint for a profile to make it clearer what the tags are for

api/TimeAddressableMediaStore.yaml
johnbilt: Removed

api/TimeAddressableMediaStore.yaml
johnbilt: As above, removed!

api/schemas/profile.json
johnbilt: Talked with Sam - we're going to remove the urn and assume that UUID's are globally unique.  This is inline with what we do with flows and sources.  Can re-visit in the future if necessary

docs/appnotes/00xx-using-flow-profiles.md
samdbmg: We had some discussion about this, and concluded that for the time being, we should leave greedy matching out for now, because it is better to be explicit than implicit. Let's revisit this if it comes up again.

api/examples/profile-post-201.json
samdbmg: It's been pointed out to me that it saves us a problem later if they need to differ, so let's leave this one

api/TimeAddressableMediaStore.yaml
johnbilt: Updated in next update....

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+If the content to be referenced is only coming from a one source (potentially with multiple flows), for example a simple clip, then it is relatively easy to clone the technical parameters of the existing flows and then join the segments to the new flow.
+However if the content for the new edit is derived from multiple sources then there is a need to match the flows from the different sources based on the technical characteristics.
+Currently the TAMS API does not provide an easy method to do this, so it is necessary to compare all the flows in code to find the matches.
+The more sources involved in this process, the harder this becomes.
+
+## TAMS Profiles within the API
+
+The TAMS profile model is split into a number of stages:
+
+* An end-point to list and describe the available profiles supported by the store
```
danjhd: nit: endpoint (no hyphenation)

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+### Creating a flow using a profile
+
+When creating a flow the generating system has two options:
+
+1. Specify all the technical characteristics of the flow including the wrapper, codec and essence parameters alongside the non-technical parameters such as label and description
+2. Provide just the technical profile for the flow and non-technical parameters required
+
+When using the second option on submission of the create flow request, the store will then read the metadata for a given profile and use this to populate the metadata for the flow.
+This de-normalisation of the technical parameters means that on the read side the flows created via both mechanisms have the same technical parameters available.
+Additionally if the decision is made to change a standard profile in the future then this does not affect the existing flows as the parameters have already been replicated and should remain unchanged to reflect the media actually being stored.
```
danjhd: Above you mention, quite rightly i believe, that Flows should be immutable. Yet here you talk about a use case where a standard profile could be changed. This conflicts. I don't believe we should allow this use case. As you say above if the profile needs to change a new one with a new Id should be created and used.

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+This would enable implementations to store additional metadata such as encoding parameters alongside the TAMS profile metadata
+
+### Creating a flow using a profile
+
+When creating a flow the generating system has two options:
+
+1. Specify all the technical characteristics of the flow including the wrapper, codec and essence parameters alongside the non-technical parameters such as label and description
+2. Provide just the technical profile for the flow and non-technical parameters required
+
+When using the second option on submission of the create flow request, the store will then read the metadata for a given profile and use this to populate the metadata for the flow.
```
danjhd: It feels here like we are making a decision for the implementation? As long as the GET request to the Flow returns all the metadata it should not matter how the implementation does this. i.e. it might not populate them it might just keep the reference to the profile id and "hydrate" on request? The important point is that when GETting Flows a client should be able to see all the metadata regardless of how it was created.

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Profiles endpoint
+
+It is possible to create and view profiles via the dedicated API calls under the TAMS service end point (/services/profiles).
+These profiles hold all the technical parameters required to create a flow in a single location.
+A system looking to ingest standardised content into TAMS would create content matching one or more profiles as defined in the endpoint.
+
+A profile should be treated as immutable, so once created it cannot be updated.
+Updating a profile would cause mismatches with flows which have been already created using that profile and so breaks the model for profiles.
+To update a profile a new one with a new ID should be created.
```
danjhd: Do we (or should we) have an option to "archive" an "old" Profile that should no longer be used but is in use already and therefore cannot easily be deleted?

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. Specify all the technical characteristics of the flow including the wrapper, codec and essence parameters alongside the non-technical parameters such as label and description
+2. Provide just the technical profile for the flow and non-technical parameters required
+
+When using the second option on submission of the create flow request, the store will then read the metadata for a given profile and use this to populate the metadata for the flow.
+This de-normalisation of the technical parameters means that on the read side the flows created via both mechanisms have the same technical parameters available.
+Additionally if the decision is made to change a standard profile in the future then this does not affect the existing flows as the parameters have already been replicated and should remain unchanged to reflect the media actually being stored.
+
+Flows that have been created from a profile will include the parameter indicating which profile they were created from.
+This differentiates them from the flows that have been created with the technical characteristics directly.
```
danjhd: Do we need to differentiate between flows that were created using a Profile and a Flow created using exactly the same Metadata as an existing Profile?
If we didn't then perhaps the store should detect this at creation time and set the Profile Id to match even if not supplied?
OR it could reject the request saying that this Flow "metadata" matches an existing Profile and therefore the profile should be used?
Just thinking out loud here.

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+
+The get flows endpoint has the ability to query for flows with a given profile.  
+When combined with other query parameters such as source id then this means it is easy to start matching content formats.
+
+For example on an edit by reference workflow it would be possible to read the profile endpoint for the recommended profiles that content should have.
+This then simplifies the process of creating the required flows and sources for the new content item.
+The edit process could then process each input source in turn read the available flows and match them to the destination flows using the profile tag easily.
+If the source content has additional non standard flows then these could be ignored.
+
+From a flow created from a profile, it is possible to query via both profile ID and also the individual parameter of the flow that have been inheritied from the profile.
```
danjhd: We should define the behaviour if these conflict. I.e. You have a profile A that defines codec as X. You query for Flows that use Profile A using the new query parameter but also use the existing codec query parameter and set that to Y. What should the behaviour be?

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+
+This document describes how flow profiles can be used within the TAMS API to simplify workflows while maintaining the flexibility of the core API.
+
+## Challenges
+
+### House formats
+
+Within a customer deployment of TAMS it is expected that there would be limited number of recommended formats for users to work in.
+Typically this could include a target format for the high quality media and standardised formats for proxy media and images.
+These formats are typically refered to as House formats and do not preclude the storing of other content within the store, however it is likely to be normalised at some point to the house format.
```
danjhd: Spelling: referred

docs/appnotes/0019-using-flow-profiles.md
danjhd: Spelling: inherited

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Multi-store working with profiles
+
+The UUID of a profile is assumed to be globally unique.
+This is the same model as for flow, source and object ID's which should be preserved when replicating content.
+
+For workflows where replication of the same content formats are happening on a regular basis then it is recommended that the same profile is loaded into both stores using the same UUID.
+This will mean than when flows are replicated between the stores then the profile identifier will continue to link to the metadata.
+
+If the profile does not existing within the destination store then the profile ID should continue to be preserved.
```
danjhd: Typo: exist

docs/appnotes/0019-using-flow-profiles.md
```diff
@@ -0,0 +1,90 @@
...
+This will continue to allow the matching of content within the store by profile ID, plus the profile can be added later and will link to the existing content.
+
+For workflows including more than two orgnaisations it is recommended that one organisation takes responsibility for owning and publishing the profiles.
```
danjhd: Spelling: organisations

danjhd (commented):
Also this may be a nit and BBC R&D can guide but my feeling is that when referring to TAMS entity types like Flow(s). Source(s) or Profile(s) we should always capitalise these words to help clarify?

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Done

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Done

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Done

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Done

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Done

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Good point - we should remove this as not required.  Will re-work this section anyway based on another comment

docs/appnotes/0019-using-flow-profiles.md
johnbilt: This probably needs the outcome of the greedy match discussion to complete.  If we're not going to implement greedy match at any level then the only way the profile_id field should be populated is if it's provided at creation.

If we want to do match on creation and populate the profile field then yes we may want to indicate this.  However the question is what use is this in the actual TAMS metadata or whether it should be captured in an observability layer for reporting/tracking.

I would probably stay away from rejecting requests where they exactly match a profile.  I would either match on creation and populate the field or leave creators the choice to use profiles or not as they wish.

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Re-worked section to hopefully including this

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Re-worked section to hopefully including this

docs/appnotes/0019-using-flow-profiles.md
johnbilt: Currently this is not in scope of this PR.  We do need to think about old profiles, however we should also consider the same approach for the storage backend endpoint under the services section also.  So previously proposed that we handle both of these at the same time in a future PR

j616: For completeness, one option is to add fine-grained auth to profiles (and storage backends) to allow their use to be restricted as required.

docs/appnotes/0020-using-flow-profiles.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Multi-store working with Profiles
+
+The UUID of a Profile is assumed to be globally unique.
+This is the same model as for Flow, Source and Object ID's which should be preserved when replicating content.
+
+For workflows where replication of the same content formats are happening on a regular basis then it is recommended that the same Profile is loaded into both stores using the same UUID.
+This will mean than when Flows are replicated between the stores then the Profile identifier will continue to link to the metadata.
+
+If the Profile does not exist within the destination store then the Profile ID should continue to be preserved.
```
j616: Do we need to describe how this interacts with the Profiles endpoint? Should the Profile be created if it doesn't exist? If not, how does it interact with the creation of a Profile with a matching ID? I think the paragraph below covers some of this. It might help if that is moved above this para?

docs/appnotes/0020-using-flow-profiles.md
```diff
@@ -0,0 +1,93 @@
...
+This will continue to allow the matching of content within the store by Profile ID, plus the Profile can be added later and will link to the existing content.
+
+For workflows including more than two organisations it is recommended that one organisation takes responsibility for owning and publishing the Profiles.
```
j616: Should this be `one organisation takes responsibility for owning and publishing a given Profile` or `shared Profiles`? I read this as all Profiles, which I don't think is practical/possible.

docs/adr/0047-flow-profiles.md
```diff
@@ -0,0 +1,68 @@
...
+### Option 2: Add Flow Profiles with advanced search criteria
+
+This option builds on Option 1 and adds an additional query parameter to define the Profile match mode on the Flows listing endpoint.
+Options would include matching flows which share the technical characteristics as a Profile but have not been associated with it.
+Alternatively it could be used as an exclude to look for all Flows which match a profile but have not been linked to it.
+
+- Good: Enables brownfield adoption — existing Flows that match a Profile's technical characteristics become discoverable
+- Good: Simplifies edit-by-reference workflows across mixed content where some Flows were created with Profiles and others were not
+- Good: Makes Profiles useful as a query template even in stores where Profile-based creation has not been universally adopted
+- Bad: Requires clear definition of what constitutes a "match"
```
j616: I might be missing something. I had assumed that all parameters specified in the Profile would have to match exactly in Flows. Or is it items like tags and other dictionaries where behaviour might have to differ?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1106,6 +1292,13 @@ paths:
...
         It may also be possible for service implementations to detect some instances where multiple Flows should not be considered of the same Source, such as audio Flows with different numbers of tracks.
         Further guidance on when Flows/Sources may be considered the same/different may be found in the [Practical Guidance for Media](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.0._Practical_Guidance_for_Media.html) section of AMWA MS-04.
+
+        Flows can be created either by directly supplying all the required technical metadata or by referencing an existing Profile in the store.  
+        When using a Profile, the store will use the technical details (format, codec, essence_parameters etc.) from that profile, along with the metadata provided.  
+        It is up to the store implementation whether this is normalised on the flow creation or read-only.
+
+        When creating a flow using a Profile it is not possible to override any specific fields as this would invalidate the link to the Profile.  
+        Supplying a Profile ID and technical metadata should result in a 400 validation error
```
j616:
```suggestion
        Supplying a Profile ID and technical metadata defined within the specified Profile should result in a 400 validation error
```

api/TimeAddressableMediaStore.yaml
j616, line 1364: Do we need to add conflicting Profiles, and invalid Profile IDs to this return code?

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,181 @@ paths:
...
+                  summary: Image Profile
+                  externalValue: examples/profile-get-200-image.json
+        "404":
+          description: The requested profile does not exist.
+    post:
+      summary: Create Profile
+      description: |
+        Create a new Profile to be used when creating flows with the matching metadata.  
+        
+        It is not possible to update an existing Profile as this would mean it no longer matches flows that have already been created using it.
```
j616: I wonder if normative language would be useful here?
```suggestion
        Services MUST reject requests that would update an existing Profile as this would mean it no longer matches flows that have already been created using it.
```

api/examples/profiles-get-200.json
```diff
@@ -0,0 +1,62 @@
+[
+  {
+    "id": "329b98d6-adeb-418c-8eaf-eca14edc1706",
+    "label": "1080p50 TS Video 50Mpbs",
```
j616: NIT: We recently added diverse formats to examples throughout the spec to help with the misconception that TAMS is TS-centric. It might be useful to add diverse formats to the profile examples too?

api/schemas/flow-put.json
```diff
@@ -0,0 +1,33 @@
...
+      "oneOf": [
+        {
+          "type": "object",
+          "title": "Flow Profile",
+          "required": [
+              "profile_id"
+          ],
+          "properties": {
+              "profile_id": {
+                  "description": "Profile identifier that was used to create the flow.  When supplying a profile_id no technical metadata should also be provided.  Doing so will result in a 400 validation error",
```
j616: The Profile schema seems to allow for partial technical metadata. I think I'd expected Flows to be able to add additional metadata not defined in the Profile. As it stands, I think the main places where metadata might diverge is captured in `flow-common.json`, so maybe thats ok? Though I don't think we've previously defined what is "technical metadata" and what is "non-technical metadata" anywhere. I feel like we might need to either generalise the approach or specify the distinction.

docs/appnotes/0020-using-flow-profiles.md
johnbilt: Updated to `a given profile`

api/TimeAddressableMediaStore.yaml
johnbilt: The line above clearly says it is not possible to override the technical parameters of a profile, so the thinking was you supply a profile or technical metadata not both.  So possibly might be better saying `Supplying a Profile ID and any technical metadata should result in a 400 validation error`

api/schemas/flow-put.json
johnbilt: Updated wording to `When supplying a profile_id no metadata which can be contained in a Profile should also be provided`

api/examples/profiles-get-200.json
johnbilt: Done

docs/adr/0047-flow-profiles.md
johnbilt: I think this is actually covered in the next line that all the possible match modes for greedy match need clear definition.  However next line called this out, so have removed this one

docs/appnotes/0020-using-flow-profiles.md
johnbilt: I think there is a larger issue here which I'd not spotted.  To create a flow you provide either a profile or the full flow metadata.  If the profile does not exist in the store then you can't actually create the flow as there is no where to get the technical metadata from.  So I think you either have to fail the creation and then the calling system either needs to create the profile and then re-try and flow creation or drop the profile and just create the flow directly.  The first should be the preferred option as it retains the profile information.

api/TimeAddressableMediaStore.yaml
j616: Sorry, yes. I had it in my head that a partial profile could be used. But I think that's a future iteration. I think your tweaked message helps. Have we explicitly defined what constitutes "technical metadata" anywhere? Specifically in the human readable version of the spec.

api/schemas/flow-audio.json
j616, line 3: When rendered, these descriptions end up as the `flow_metadata` description for each type at the `/profiles` endpoint. I think this is unintended given the specific description in the profile schema

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,184 @@ paths:
...
+      description: Return Profile headers
+      operationId: HEAD_profiles
+      tags:
+        - Profiles
+      parameters:
+        - name: format
+          in: query
+          description: Filter on Profile format.
+          schema:
+            $ref: 'schemas/content-format.json'
```
j616: This includes multi, though multi-profiles aren't possible

api/TimeAddressableMediaStore.yaml
```diff
@@ -146,6 +146,184 @@ paths:
...
+        content:
+          application/json:
+            example:
+              $ref: examples/profile-put.json
+            schema:
+              $ref: schemas/profile.json
+        required: true
+      responses:
+        "201":
+          description: The flow has been created.
```
j616:
```suggestion
          description: The Profile has been created.
```

api/schemas/flow-technical.json
```diff
@@ -0,0 +1,61 @@
...
+            "type": "string",
+            "pattern": "^[^\\s\/]+/[^\\s\/]+$"
+        },
+        "container": 
+        {
+            "description": "The container MIME type for Flow Segments. Note that the `type` component of the container MIME type (i.e. the component before the `/`) may be different to the `type` component of the codec MIME type. e.g. An audio Flow may have `audio/aac` coded content may be wrapped in a `video/mp2t` container. Where multiple types exist for a subtype (e.g. `video/mp4`, `audio/mp4`, `application/mp4`), the closest MIME type to the Flow `format` should be used (e.g. `audio/mp4` for a Flow `format` of `urn:x-nmos:format:audio`). Mime types from the [IANA registry](https://www.iana.org/assignments/media-types/media-types.xhtml) should be preferred. Where multiple MIME types are possible, the most common should be preferred. Where this is insufficient, the maintainers of the TAMS repository may create an application note advising which MIME type to use.",
+            "type": "string",
+            "pattern": "^[^\\s\/]+/[^\\s\/]+$"
+        },
+        "avg_bit_rate":
```
j616: I think `avg_bit_rate` and `max_bit_rate` either need re-locating such that they aren't in the profile, or we need to add notes to their specific CRUD endpoints stating they can't be modified where profile is set. iirc, they are update-able because they can't always be known when the Flow ingest is configured (e.g. when configuring a push-based live ingest sink) and they may vary over time in some cases.

api/schemas/flow-get.json
```diff
@@ -0,0 +1,30 @@
...
+    "title": "Flow",
+    "description": "Describes a Flow",
+    "type": "object",
+    "properties": 
+    {
+        "profile_id": 
+        {
+            "description": "Profile identifier that was used to create the flow.",
+            "type": "string",
+            "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
```
j616:
```suggestion
            "$ref": "uuid.json"
```

api/schemas/flow-put.json
```diff
@@ -0,0 +1,33 @@
...
+          "type": "object",
+          "title": "Flow Profile",
+          "required": [
+              "profile_id"
+          ],
+          "properties": {
+              "profile_id": {
+                  "description": "Profile identifier that was used to create the flow.  When supplying a profile_id no metadata which can be contained in a Profile should also be provided, doing so will result in a 400 validation error.  Trying to create a Flow using a Profile ID that does not exist should also provide a 400 validation error",
+                  "type": "string",
+                  "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
```
j616:
```suggestion
                  "$ref": "uuid.json"
```

api/schemas/profile.json
```diff
@@ -0,0 +1,57 @@
...
+    "id",
+    "flow_metadata"
+  ],
+  "properties": 
+  {
+    "id": 
+    {
+      "description": "Profile identifier.",
+      "type": "string",
+      "pattern": "^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$"
```
j616:
```suggestion
      "$ref": "uuid.json"
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1106,6 +1295,13 @@ paths:
         Examples of conflicting metadata include `format` not matching, or the `role` in `source_collection` and `flow_collection` not matching.
         It may also be possible for service implementations to detect some instances where multiple Flows should not be considered of the same Source, such as audio Flows with different numbers of tracks.
         Further guidance on when Flows/Sources may be considered the same/different may be found in the [Practical Guidance for Media](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.0._Practical_Guidance_for_Media.html) section of AMWA MS-04.
+
+        Flows can be created either by directly supplying all the required technical metadata or by referencing an existing Profile in the store.  
+        When using a Profile, the store will use the technical details (format, codec, essence_parameters etc.) from that profile, along with the metadata provided.
```
j616:
```suggestion
        When using a Profile, the Service Implementation will populate the Flow's technical parameters (format, codec, essence_parameters etc.) from that profile, along with the metadata provided.  
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1106,6 +1295,13 @@ paths:
         Examples of conflicting metadata include `format` not matching, or the `role` in `source_collection` and `flow_collection` not matching.
         It may also be possible for service implementations to detect some instances where multiple Flows should not be considered of the same Source, such as audio Flows with different numbers of tracks.
         Further guidance on when Flows/Sources may be considered the same/different may be found in the [Practical Guidance for Media](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.0._Practical_Guidance_for_Media.html) section of AMWA MS-04.
+
+        Flows can be created either by directly supplying all the required technical metadata or by referencing an existing Profile in the store.  
+        When using a Profile, the store will use the technical details (format, codec, essence_parameters etc.) from that profile, along with the metadata provided.  
+        It is up to the store implementation whether this is normalised on the flow creation or read-only.
```
j616:
```suggestion
        It is up to the Service Implementation whether this is normalised on the Flow creation or on read.
```
Tweaks based on convention for referring to TAMS entities. Slight tweak to `read-only` due to that being a specific parameter.

api/TimeAddressableMediaStore.yaml
johnbilt: Updated to move to inline schema as suggested and multi removed.  Please double check you are happy with the update

api/schemas/flow-technical.json
johnbilt: For tracking purposes, after discussion the plan is as follows:

1. `max_bit_rate` will move back to being in the flow as this is something that should be measured at the point of content creation and since it's not fixed should not be in the profile

2. `avg_bit_rate` will remain in the profile as in practise this should be the same as the value the encoder is using as it's target bitrate.  Also for profiles to work you need to specify the target bitrate to be able to distinguish between different bitrate versions of the same codec (eg 50Mbps v 100Mbps).  This change does effectively move average to being an expectation, which is how it's really being used and achieves the aim of the field, so the ADR which talks about this is going to need updating

3. Wording needs to be added around what happens if someone tries to update a field in a flow which has come from a profile.  It is expected that to do this the calling system will remove the `profile_id` by leaving it blank which will break the link to the profile as part of update message.  Updates where the `profile_id` is still present SHOULD be rejected by the store as invalid with a suitable error message.

docs/appnotes/0020-using-flow-profiles.md
```diff
@@ -77,6 +77,21 @@ As per standard behaviour, the API should only return results which match all fi
 In future it is proposed to add additional match methods when querying Flows using a Profile.
 This would include options for a "greedy match" where all Flows which match the technical characteristics are returned even if not created or updated to include a Profile.
 
+### Updating Flows created using a Profile
+
+For Flows created by specifying all technical characteristics directly, updates can include any of those parameters without restriction. 
+When a Flow is created from a Profile then it inherits all the technical characteristics from the Profile.
+This means if that Flow then needs to be updated then consideration needs to be given to the fields which were inherited.
+
+For the standard fields (eg label, description, tags) which form part of the common Flow metadata and provided directly regardless of the Profile then these can be updated through the standard process.
```
j616, line 86: I think this and the following paragraph need to be moved/copied to the top level Flow PUT endpoint's top level description (the one in the main YAML file) and the average bitrate PUT as it describes normative behaviour Service implementations should implement and Clients should expect. AppNotes are usually considered Informative/advisory.

j616 (changes requested):

api/TimeAddressableMediaStore.yaml
j616: Looks good

api/schemas/flow-technical.json
samdbmg: I've put in https://github.com/bbc/tams/issues/236 to consider making the bitrate thing clearer overall in the spec

api/schemas/flow-audio.json
j616: It's a little ugly, but the best way I could find to make this work is to move the `description` under the items in `oneOf`. I think it's because the `description` is tied to the selectable items, rather than the top level parameter.

johnbilt: Have added descriptions up one level into flow-get.json and flow-put.json.  Hoping this will resolve the rendering issues

api/TimeAddressableMediaStore.yaml
```diff
@@ -1106,6 +1303,21 @@ paths:
...
+        It is up to the Service Implementation whether this is normalised on the Flow creation or on read.
+
+        When creating a flow using a Profile it is not possible to override any specific fields as this would invalidate the link to the Profile.  
+        Supplying a Profile ID and any technical metadata should result in a 400 validation error.
+
+        When updating a Flow created using a profile then  consideration needs to be given to the fields which were inherited. 
+        For the standard fields (eg label, description, tags) which form part of the common Flow metadata and provided directly regardless of the Profile then these can be updated through the standard process.
+
+        In the scenario where the technical characteristics do need to be changed on a Flow created from a Profile then it is necessary to remove the link to the Profile as part of the update since it will no longer match the Profile entirely.
+        This is achieved by means of sending the `profile_id` field with an empty string as part of the update.
```
j616:
```suggestion
        This is achieved by means of setting the `profile_id` field to an empty string as part of the update.
```
Slight tweak. "Sending" didn't want to parse properly in my brain. But happy to leave it as is if it's just me 🙂

api/TimeAddressableMediaStore.yaml
```diff
@@ -1686,6 +1901,11 @@ paths:
 
         The average bit rate of the Flow Segments in 1000 bits/second.
         A precise definition can be found in the [Setting Flow Bit Rate Properties](https://github.com/bbc/tams/blob/main/docs/appnotes/0013-setting-flow-bit-rate-properties.md) AppNote.
+
+        When a Flow is created from a Profile then the average bit rate is inherited from the Profile.
+        In this scenario is it not possible to update the avrage bit rate directly as it would invalidate the link to the Profile.
```
j616:
```suggestion
        In this scenario is it not possible to update the average bit rate directly as it would invalidate the link to the Profile.
```

docs/appnotes/0020-using-flow-profiles.md
johnbilt: Added into the main spec YAML file in both places

api/schemas/flow-audio.json
johnbilt: Removed and move to the correct place in schemas/profile.json.  @j616 confirmed now renders correctly

j616 (approved):
Changes look good 💯

PR #134, opened 2025-06-17
Bugfix: Remove container mapping from source collections

j616:
# Details
https://github.com/bbc/tams/pull/46 added container mapping support to TAMS. It was intended that this would only apply to Flows, but was accidentally added via the collection schema which was included in the source metadata. This resulted in this technical metadata being included in Source Collections. Technical metadata should only be present in Flows, not Sources. This bug has been present since TAMS version [5.0](https://github.com/bbc/tams/releases/tag/5.0). This PR fixes this.

Note that the spec states the following for Source Collections:
> This attribute is intended to be read-only. Implementations SHOULD ignore this if given in a PUT request, and instead manage it internally. Source collections can be inferred from Flow collection definitions.

As such this change may impact server implementations which previously propagated this data, and clients which may read these parameters. It should not impact the writing of data.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5449

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/.github/workflows/adr-check-workflow.yml
+++ b/.github/workflows/adr-check-workflow.yml
@@ -1,0 +1,46 @@
+name: ADR and AppNote Check
+run-name: ${{ format('{0} triggered by {1} on {2} {3}', github.workflow, (github.event_name == 'workflow_dispatch' && format('user {0}', github.actor) || format('{0} event', github.event_name) ), github.repository, github.ref_name) }}
+on:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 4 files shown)


GeorginaShippey (approved):
Looks good to me!

PR #9, opened 2023-12-19
Make timerange more prominent in docs

samdbmg:
# Details
Moves the definition of TimeRange and Timestamp into a JSON schema file, and references it in all the types that use it.
Also adds links from the descriptions in rendered docs.

# Pivotal Story
Story URL: https://www.pivotaltracker.com/story/show/186709535

# Related PRs
Follows https://github.com/bbc/tams/pull/8

# Links to external test runs/working deployment
N/A

# Submitter PR Checks
_(tick as appropriate)_

- [X] Added bbc/rd-apmm-cloudfit team as a reviewer
- [X] PR completes task/fixes bug
- [ ] Tests exercise code appropriately
- [ ] New features and API breaks are flagged in commit messages using magic strings
- [ ] Repo maintainer is notified that a release is required
- [X] Documentation updated (README, Confluence, Docstrings, API spec, Engineering Guide, etc.)
- [ ] Downstream repos have been checked for potential breaks & fixed as needed
- [X] APIs/UIs/CLIs updated as required
- [X] PR added to Pivotal story
- [ ] Follow-up stories added to Pivotal
- [ ] Any temporary code/configuration removed (e.g. test deployment environment, temporary commontooling branch)
- [ ] Any pins against pre-releases have been removed

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] PR completes task/fixes bug
- [ ] Tests exercise code appropriately
- [ ] Design makes sense, and fits with our current code base
- [ ] Code is easy to follow
- [ ] PR size is sensible
- [ ] Commit history is sensible and tidy

# Info on Cloudfit PRs
- The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.
- For more details on how to asses PRs, see: https://github.com/bbc/rd-apmm-docs-ways-of-working/blob/master/workflow/PRChecklist.md

PR #140, opened 2025-07-09
Allow `ts_offset` to be a negative offset

samdbmg:
# Details
Relaxes an erroneous restriction that `ts_offset` cannot be negative: the offset maps the media object timeline into the Flow timeline, and there is no particular stipulation as to which way that offset occurs.

~~*Question*: Are we happy this is a fix since it seems obvious it's wrong, or does it deserve a version bump?~~ - see comment below

(thanks to @johnbilt for the report)

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5459

# Related PRs
Introduced by https://github.com/bbc/tams/pull/9

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/schemas/flow-segment-post.json
+++ b/api/schemas/flow-segment-post.json
@@ -12,7 +12,7 @@
       "type": "string"
     },
     "ts_offset": {
-      "description": "The timestamp offset between the sample timestamps stored in the media file and the corresponding timestamp in the segment, ie. ts_offset = segment ts - media object ts. Assumed to be 0:0 if not set. Format as described by the [Timestamp](../schemas/timestamp#top) type, but cannot be negative",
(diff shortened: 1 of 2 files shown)


GeorginaShippey (dismissed):

samdbmg:
> Question: Are we happy this is a fix since it seems obvious it's wrong, or does it deserve a version bump?

We had a chat on Slack about this, and we are. Notably there are Flows in stores used by CNAP powered by https://github.com/awslabs/time-addressable-media-store that have negative offsets (so the rule isn't enforced there) and it makes things more permissive rather than less.

PR #141, opened 2025-07-18
Better explain how to work out which parts of a media object to use

samdbmg:
# Details
- Expands the description of how to choose which samples inside a media object should be used for a Flow Segment, when only part of that object is being used
- Clarifies that segments have to be internally consistent between `timerange` and `sample_offset`/`sample_count`

**Note:** ADR moved to separate PR https://github.com/bbc/tams/pull/147  ~ADR added to aid decision process~ ~Draft for now to aid with discussion: it's possible keeping offset and count doesn't make sense. After discussion this will be the topic of an ADR)~ 

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5400

# Related PRs
https://github.com/bbc/tams/pull/147 considers whether to remove offset and count entirely

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


README.md
```diff
@@ -138,6 +138,25 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
+Alternatively if `sample_offset` and `sample_count` are set on the segment, a client may count the samples read from the object and discard accordingly, although care must be taken with temporal re-ordering, since the `sample_offset` and `sample_count` refer to the presentation timeline, rather than the decode timeline.
+
+![Graphic showing the objects making up Flow Z (above) and their segments, along with the selected grains](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ-subsegments.drawio.png)
+
+The diagram above shows a smaller portion of the Flow Z timeline considered previously, with the presentation time stamp (PTS) of each grain in the object along the bottom.
+The pseudocode below shows how a client might apply the `ts_offset` to each PTS, then validate whether it is inside the given `timerange`.
+
+```python
+grain_ts = pts_to_timestamp(grain_pts)             # `4:0` for `grain_pts = 40`
+grain_ts_in_flow_timeline = grain_ts + ts_offset   # `4:0 + -0:700000000` = `3:300000000`
```
j616: NIT: The placement of quote marks on this comment seems a little odd
```suggestion
grain_ts_in_flow_timeline = grain_ts + ts_offset   # `4:0 + -0:700000000 = 3:300000000`
```

samdbmg: Good spot, thanks!

samdbmg:
After various discussions, I've realised talking about PTS isn't entirely helpful here, because it's very MPEG-specific. I'll tweak the terminology as needed, and also I have a more thorough diagram to include

api/schemas/flow-segment.json
```diff
@@ -24,11 +24,11 @@
       "$ref": "timestamp.json"
     },
     "sample_offset": {
-      "description": "The start of the segment represented as a count of samples from the start of the object. Note that a sample is a video frame or audio sample. A (coded) audio frame has multiple audio samples. Assumed to be 0 if not set.",
+      "description": "The start of the segment represented as a count of samples from the start of the object. Note that a sample is a video frame or audio sample. A (coded) audio frame has multiple audio samples. Assumed to be 0 if not set. Must be set if the Flow Segment doesn't start at the beginning of the media object.",
```
j616, line 27: I guess we aren't considering this a breaking change because its a clarification of "Assumed to be 0 if not set"? i.e. if you weren't setting this and `sample_offset` isn't 0, then the metadata is wrong.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1645,14 +1645,14 @@ paths:
         "http_object_store" media store type which MUST include a `get_urls` property that contains the
         HTTP URLs for downloading the media object - server implementations should generate this internally.
 
-        The flow segment may include timing and timerange adjustment information that the client needs to
-        apply when extracting the samples from the media object.
-        - If `sample_count` is set, the timerange of samples must be `sample_offset` to
-         `sample_offset + sample_count` (exclusive). Otherwise, the timerange of samples must be
-          `sample_offset` until the end.
+        The flow segment may include timing adjustment information that the client needs to apply when extracting
+        the samples from the media object, and may not use the entirety of the samples in the media object.
```
j616: I don't think this flows quite right. Is this saying that the flow segment might not use the entirety of the media object? Or that the adjustment information is included when the entirety of the object isn't used?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1645,14 +1645,14 @@ paths:
...
         - The sample timestamp in the flow segment (`segment_ts`) is the timestamp in the media object
```
j616:
```suggestion
        - The timestamp of a sample on the flow segment's timeline (`segment_ts`) is the timestamp of that sample embedded in the media object, or otherwise inferred from the media object,
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1645,14 +1645,14 @@ paths:
...
-          (`media_object_ts`) adjusted by `ts_offset`: `segment_ts = media_object_ts + ts_offset`. The
-          `segment_ts` should equal the start of the `timerange` for the sample at `sample_offset`.
+          (`media_object_ts`) adjusted by `ts_offset`: `segment_ts = media_object_ts + ts_offset`.
+        - If the segment only uses part of the object, that part is where `segment_ts` falls within the segment
```
j616: I'm not quite sure what this bullet point is trying to explain. I think it's trying to say any given `media_object_ts` is only used where it's calculated `segment_ts` is contained in `timerange`? I'm not sure we need to call that out if we get the above mapping explanation right?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1645,14 +1645,14 @@ paths:
...
+          `timerange`
+        - In addition, the samples within `timerange` must start after `sample_offset` samples on the presentation
```
j616: The current wording implies that the timerange and sample offset/count don't have to be precisely align.
```suggestion
        - In addition, the segment's `timerange` start MUST be at `sample_offset` samples from the start of the presentation
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1645,14 +1645,14 @@ paths:
...
+          timeline of the object, and end after `sample_offset + sample_count` samples.
```
j616:
```suggestion
          timeline of the object, and `timerange` end MUST be at `sample_offset + sample_count` samples.
```

README.md
```diff
@@ -112,7 +112,7 @@ Various scenarios are explored in the [Practical Guidance for Media](https://spe
 
 Flows exist on an infinite timeline (the "Flow timeline"), and the position of content on this timeline is defined by the `timerange` attribute in each Flow Segment of that Flow.
 A timerange is represented in JSON and text using the [TimeRange string pattern](https://bbc.github.io/tams/5.1/index.html#/schemas/timerange).
-Separately the media objects have a timeline (the "media timeline") defined by the container format itself: the timestamps recorded inside the media object for each grain.
+Separately the media objects have a timeline (the "media timeline") defined by the container format itself: the timestamps recorded inside the media object for each grain in the natural format for that container (e.g. the PTS timing for MPEG-TS).
```
j616:
```suggestion
Separately the media objects have a timeline (the "media timeline") defined by, or implied by, the container format itself: the timestamps recorded inside the media object for each grain in the natural format for that container (e.g. the PTS timing for MPEG-TS, or the time from the start of the media object for v210 of a given frame rate).
```

README.md
```diff
@@ -138,6 +138,30 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
 
 ![Graphic showing the Flow timeline and Flow Segments of Flows X, Y and Z, where Z is composed of a mix of re-used segments and new media](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ.drawio.png)
 
+In practice a client may need to read (and potentially decode, in the case of inter-frame video codecs) the entire object, discarding the unwanted grains and returning those selected by the segment to the consumer.
+This can be handled by mapping the object's timeline into the Flow timeline by adding `ts_offset` to each timestamp inside the media object, and then discarding grains when the resulting timestamp falls outside the given `timerange` for the segment.
+Alternatively if `sample_offset` and `sample_count` are set on the segment, a client may count the samples read from the object and discard accordingly, although care must be taken with temporal re-ordering, since the `sample_offset` and `sample_count` refer to the presentation timeline, rather than the decode timeline.
```
j616:
```suggestion
Alternatively if `sample_offset` and `sample_count` are set on the segment, a client may count the samples read from the object and discard accordingly, although care must be taken with codecs that use temporal re-ordering, since the `sample_offset` and `sample_count` refer to the presentation timeline, rather than the decode timeline.
```

README.md
```diff
@@ -138,6 +138,30 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
+
+![Graphic showing the objects making up Flow Z (above) and their segments, along with the selected grains](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ-subsegments.drawio.png)
+
+The diagram above shows three of the objects used by Flow Z and a smaller portion of the Flow Z timeline considered previously, with both the media object timeline (and `media_ts`) and Flow timeline (`grain_ts`) of each grain in the object along the bottom.
```
j616: Is it confusing to use `grain_ts` here and `segment_ts` in the API spec?

docs/images/Flow and Media Timelines-Flow XYZ-subsegments.drawio.png
j616, line 1: I wonder if this should say `media object timelines` or some other plural form? Mainly because each object's `media_ts`'s might be on a different timeline. Maybe they shouldn't be presented on the same horizontal line at all? Alternatively, should the `grain_ts` row use a single colour along its length to indicate its the same timeline?

README.md
```diff
@@ -138,6 +138,30 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
+The `timerange`/`ts_offset` approach is illustrated by the following pseudocode, which shows how a client might apply the `ts_offset` to each `media_ts`, then validate whether it is inside the given `timerange`.
+Note that the `media_ts` may be a different precision or timebase to the nanosecond timestamps used by TAMS, so some conversion may be required.
+
+```python
+grain_ts = media_ts_to_timestamp(media_ts)         # `4:0` for `media_ts = 4.0`
```
j616: Add units?
```suggestion
grain_ts = media_ts_to_timestamp(media_ts)         # `4:0` for `media_ts = 4.0sec`
```

j616: Also, should the variable names be different here? You defined `grain_ts` in the body of text above as being on the Flow timeline.

README.md
```diff
@@ -138,6 +138,30 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
+grain_ts_in_flow_timeline = grain_ts + ts_offset   # `4:0 + -0:700000000 = 3:300000000`
+if (grain_ts_in_flow_timeline < timerange.start or
+    grain_ts_in_flow_timeline >= timerange.end):   # `3:300000000 is less than 3:500000000`
+    discard_grain()                                # So this grain is discarded
+else:
+    keep_grain()                                   # `media_ts = 4.2` will be retained, with `grain_ts = 3:500000000 (or 3.5sec)`
```
j616: Add units?

I found this a little confusing at first, btw. The change in `media_ts` from all the other comments in the snippit isn't immediately obvious. This suggestion is pretty verbose though
```suggestion
    keep_grain()                                   # Another sample of `media_ts = 4.2sec` will be retained, with `grain_ts = 3:500000000 (or 3.5sec) as 3:500000000 >= 3:500000000 (timerange.start)`
```

docs/images/Flow and Media Timelines-Flow XYZ-selecting-units.drawio.png
j616, line 1: Does this need updating to remove `pts` too?

README.md
```diff
@@ -138,6 +138,30 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
+```
+
```
j616, line 162: I wonder if its also worth providing example ffmpeg/gstreamer commands, where possible? I expect most implementations won't be dealing with individual frames themselves, and so the pseudocode won't be of much use to them.

api/schemas/flow-segment.json
samdbmg: Yes, that was my thinking - that additional sentence makes the meaning of the previous one more clear without changing it

api/TimeAddressableMediaStore.yaml
samdbmg: Yes - I was worried that it could be interpreted the other way, in that `timerange` refers to the entire segment for some reason. Do you think it's clear what `timerange` means without this bullet (because the other explanation doesn't make much sense?). I'll remove it if so

api/TimeAddressableMediaStore.yaml
samdbmg: Thanks - your wording is much better!

api/TimeAddressableMediaStore.yaml
j616: I think so. If you're concerned about a miss understanding of the use of the term `timerange` here, I think it'd be better to just re-state that definition.

api/TimeAddressableMediaStore.yaml
samdbmg: I've split it into two sentences to try and make it clearer

api/TimeAddressableMediaStore.yaml
samdbmg: Thanks - I've slightly tweaked your wording to avoid confusing the subordinate clause and the brackets

README.md
samdbmg: Thanks - I've tweaked your suggestion to avoid pre-supposing how we might deal with V210 and broken it into another sentence for flow

README.md
samdbmg: I've changed the variable names to try and be consistent with the diagrams

README.md
samdbmg: I think the change in variable names has helped here, but I've been clearer in the comment now too

docs/images/Flow and Media Timelines-Flow XYZ-subsegments.drawio.png
samdbmg: I agree on the plural form, but I think making them separate horizontal lines makes the diagram too unwieldy: instead I've gone with your colour suggestion

docs/images/Flow and Media Timelines-Flow XYZ-selecting-units.drawio.png
samdbmg: I've instead made more explicit that this example in in MPEG-TS, to try and connect the abstract timing concepts to what you'd actually find in a real file

README.md
```diff
@@ -140,25 +141,35 @@ In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z,
...
 The `timerange`/`ts_offset` approach is illustrated by the following pseudocode, which shows how a client might apply the `ts_offset` to each `media_ts`, then validate whether it is inside the given `timerange`.
 Note that the `media_ts` may be a different precision or timebase to the nanosecond timestamps used by TAMS, so some conversion may be required.
 
 ```python
-grain_ts = media_ts_to_timestamp(media_ts)         # `4:0` for `media_ts = 4.0`
-grain_ts_in_flow_timeline = grain_ts + ts_offset   # `4:0 + -0:700000000 = 3:300000000`
-if (grain_ts_in_flow_timeline < timerange.start or
-    grain_ts_in_flow_timeline >= timerange.end):   # `3:300000000 is less than 3:500000000`
-    discard_grain()                                # So this grain is discarded
+sample_ts = ts_to_timestamp(ts)      # `4:0` for `ts = 4.0sec`
```
j616: NIT: We commonly use `ts` as a shorthand for TAMS `timestamp`s. We're using that here, in fact, in the naming of variables. So maybe `ts_to_timestamp(ts)` is a little confusing? But I guess it is consistent with ffmpeg :upside_down_face:

api/schemas/flow-segment.json
j616, line 1: One more NIT: I think this needs to be reflected in `flow-segment-post.json`

README.md
```diff
@@ -112,7 +112,8 @@ Various scenarios are explored in the [Practical Guidance for Media](https://spe
 
 Flows exist on an infinite timeline (the "Flow timeline"), and the position of content on this timeline is defined by the `timerange` attribute in each Flow Segment of that Flow.
 A timerange is represented in JSON and text using the [TimeRange string pattern](https://bbc.github.io/tams/5.1/index.html#/schemas/timerange).
-Separately the media objects have a timeline (the "media timeline") defined by the container format itself: the timestamps recorded inside the media object for each grain.
+Separately the media objects have a timeline (the "media timeline") defined by, or implied by, the container format itself: the timestamps recorded inside the media object for each grain in the natural format for that container.
```
GeorginaShippey, line 115: NIT: Thinking about the other documentation PR, should we be capitalising words like Media Object. I can see a few instances where we are not sticking with this style.

j616: I was going to pick that up in my re-base :)

PR #142, opened 2025-07-25
Improvements to webhooks endpoints

j616:
# Details
While developing fine-grained auth approaches in https://github.com/bbc/tams/pull/115 , several deficiencies were noticed with the Webhooks endpoints:
* The CRUD patterns on the Webhooks endpoint do not match the rest of the API
* The use of a secret attribute for auth on this endpoint is vulnerable to the loss of that secret. This would prevent editing or deleting of a webhook.
* The BBC R&D (internal experimental) implementation and AWS implementation of TAMS differed in their interpretation of what is the primary key
* Strict interpretation of the spec, and hiding of the secret `api_key_value` in the Webhooks listing, could result in multiple identical looking Webhooks being returned
* The existing approach prevents "matrixing" of event types and filters (e.g. to receive all `flows/created` events, but only `flows/segments_added` for specific flows)

This PR proposes changes to the Webhooks endpoint to fix these issues.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5461

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


api/schemas/webhook-with-id.json
```diff
@@ -0,0 +1,23 @@
+{
+    "title": "Register Webhook",
+    "description": "Register to receive updates via webhook",
```
samdbmg: Since I think this schema only applies to responses, not requests, should these two change?
```suggestion
    "title": "Webhook Details",
    "description": "Details of an existing registered webhook",
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -170,66 +170,151 @@ paths:
...
+        format in the `webhooks` section of the API docs, depending on the event type (as defined in the same section).
+        Availability of this endpoint is indicated by the name "webhooks" appearing in the `event_stream_mechanisms`
+        list on the service endpoint.
+
+        HTTP requests from the service SHOULD include a `api_key_name` header with the 'api_key_value' value. Clients SHOULD verify this against the value they provided when registering the webhook.
+
+        API implementations MAY partially support event filtering and transformations.
+        API implementations SHALL return a 400 response code if the filtering or transformation specified in the request is not supported.
+
+        API implementations SHOULD consider the security implementations of providing webhooks, and include appropriate mitigations against Server Side Request Forgery (SSRF) attacks and similar. API implementations SHOULD take appropriate steps to authorize the modification of existing webhooks. This may take the form of RBAC, or ABAC.
```
samdbmg: I think this might need a tweak to match this being an update endpoint now

api/schemas/webhook-with-id.json
j616: Its also used in the response to registering a webhook. But I think that's fine.

api/TimeAddressableMediaStore.yaml
j616: Tweaked in https://github.com/bbc/tams/pull/142/commits/fdcd7447cc2fa2bce514f71741d3c15dc84cfbbf

api/TimeAddressableMediaStore.yaml
```diff
@@ -170,66 +170,151 @@ paths:
...
+          description: Success. The webhook has been updated
+          content:
+            application/json:
+              example:
+                $ref: "examples/webhook-get-200.json"
+              schema:
+                $ref: schemas/webhook.json
         "204":
-          description: Success. The webhook has been removed
+          description: No content. The webhook has been deleted.
```
GeorginaShippey: I'm unsure as to whether a 204 is still appropriate here given the changes.

j616: This was leftover from an earlier iteration where we specified that removing all events from a webhook should result in it being deleted. But we identified a use case where it's useful to keep the configuration, with it's secret, to effectively "pause" the events without having to re-generate the secret.

I've removed it now.

api/schemas/webhook.json
```diff
@@ -15,10 +15,6 @@
             "description": "The HTTP header name that is added to the event POST",
             "type": "string"
         },
-        "api_key_value": {
-            "description": "The value that the HTTP header 'api_key_name' will be set to",
-            "type": "string"
-        },
         "events": {
             "description": "List of event types to receive",
             "type": "array",
```
danjhd, line 20: I would like to suggest we add enum to the events array so that we can validate the event supplied is a valid webhook event. I have had people recently trying to register webhooks and have not read the docs correctly and putting incorrect strings in here. I think it would be nice to return a 400 if the event string does not match those "allowed"?

j616: Good idea! Added in https://github.com/bbc/tams/pull/142/commits/7eec78b27a67d99511b4dedb2693b0235eb4deb9

api/TimeAddressableMediaStore.yaml
```diff
@@ -168,68 +188,169 @@ paths:
...
+              $ref: "examples/webhook-post.json"
             schema:
               $ref: schemas/webhook-post.json
         required: true
       responses:
         "201":
-          description: Success. The webhook has been registered or updated
-        "204":
-          description: Success. The webhook has been removed
+          description: Success. The webhook has been updated
```
danjhd: Just wondering if this should be `description: Success. The webhook has been registered` ? If my understanding is correct of the new proposal this POST endpoint will only ever create webhooks not allow update of existing?

j616: Correct. Fixed in https://github.com/bbc/tams/pull/142/commits/352d6479351f23d3cb3f53e4d19488bf6ea70272

j616:
> Something that came up in a conversation today and may be something to include in this PR. Should we be supplying more "data" in the deleted events payload, for flows and sources? We had a use case where we wanted to be notified if a "multi" Source is deleted. Since the payload for deleted events only has the id this is not currently easy/possible.

I think that sounds reasonable. Would you be able to provide some user stories/use cases we can use to inform the change? I'm mainly wondering if that data should be the whole Source/Flow metadata, or just a subset. I wonder if it should be a separate ADR and/or PR to avoid adding even more to this very large change set!

api/TimeAddressableMediaStore.yaml
```diff
@@ -168,68 +188,169 @@ paths:
...
+        required: true
+      responses:
+        "201":
+          description: Success. The webhook has been updated
+          content:
+            application/json:
+              example:
+                $ref: "examples/webhook-get-200.json"
+              schema:
+                $ref: schemas/webhook.json
```
danjhd: Just wondering if this was a conscious decision to use `webhook` and not `webhook-with-id` as the response to the PUT? It feels like an inconsistency with other PUT methods we have in the API (e.g. Flows) that still return the id?

api/TimeAddressableMediaStore.yaml
```diff
@@ -168,68 +188,169 @@ paths:
...
+
+        API implementations SHOULD consider the security implementations of providing webhooks, and include appropriate mitigations against Server Side Request Forgery (SSRF) attacks and similar. API implementations SHOULD take appropriate steps to authorize the deleting of webhooks. This may take the form of RBAC, or ABAC.
+      operationId: DELETE_webhooks-webhookId
+      tags:
+        - Webhooks
+      responses:
+        "204":
+          description: No content. The webhook has been deleted.
+        "403":
+          description: Forbidden. You do not have permission to modify this webhook.
```
danjhd, line 354: Since we don't have a `read-only` field on the webhook, like we do on Flows i was just wondering if this 403 is needed explicitly? Since i guess this will be produced by the RBAC/ABAC/Coarse access rules? Unless i have missed something. Feels like if we are including a 403 here we should go back and add 403 to all methods?

danjhd:
I think it might be a good idea to have a `status` field (values of enabled and disabled) on a webhook. This would serve 2 purposes:

1) It would be updateable on the PUT endpoint so that a client can disabled a webhook for a period if needed without deleting it and without the need to alter it in any other way.
2) It would allow the implementation to automatically disable webhooks after a defined number of failures so avoid continuously loading the system with POSTs that will fail. The addition of an optional `error` field (readonly) would be useful to give feedback to the client on the reason for the disablement if it was done by the system rather than the client.

In our implementation i need to solve the problem of repeated failures and thought about adding a disabled flag but it think this might be useful in the main spec, and also worth adding as part of this PR as it is the same schema?

api/TimeAddressableMediaStore.yaml
j616: You're right. This should be with ID. The without ID should just be for the POST request body for the creation of new webhooks

api/TimeAddressableMediaStore.yaml
j616: Yes. This was added for access control use cases. We should make sure we're consistent with this.

api/TimeAddressableMediaStore.yaml
j616: Resolved in https://github.com/bbc/tams/pull/142/commits/18c78b23014c5d007be8c5936a548689c3e70afa

api/schemas/webhook-get.json
```diff
@@ -0,0 +1,33 @@
...
+            "required":
+            [
+                "status"
+            ],
+            "properties": {
+                "error": {
+                    "description": "Provides more information for the error status, as described by the [Error](../schemas/error#top) type",
+                    "$ref": "error.json"
+                },
+                "status": {
```
samdbmg, line 20: Sense check: are we happy this strikes the right balance of simplicity versus richness? The GitHub webhook experience is probably too far towards "richness" here: it shows you recent deliveries and their payloads and headers etc. Should we consider adding "last delivery" or "successful delivery count" or generally transforming the `error` key into more of a status? Or is this enough, and the store logs should have the rest?

samdbmg (approved):
One question, but LGTM

api/schemas/webhook-get.json
j616: This approach was taken to match the pattern used on the delete requests endpoint. The addition being that we're adding the ability to temporarily disable the webhooks, and to "re-set" them on an error condition via the `status` param. The problem with relying on store logs is that most clients won't have access to them. I'll defer to @danjhd for thoughts on "last delivery"/"successful delivery count" though. Though it would be easy enough to add that in future. So maybe this approach is enough for now?

samdbmg: Yes, I think that makes sense, and as you point out, adding more is easy enough

danjhd: My thinking was to first and foremost have a field that allows the TAMS API implementation to disable "bad" webhooks when failures are so numerous it affects performance. The idea of the "error" field was a way to pass some basic info back to the client on why it was disabled. So i suppose i had in my head that our implementation would just include the "last error" and nothing more than that. But i also feel this would be a decision by the implementation and we don't need to be specific like this in the spec we just provide a field where it can be "put"? Sorry if i'm missing the point somehow.

j616: I think that matches my thoughts

j616:
Rebasing again

api/schemas/webhook-get.json
samdbmg: Ok, I'm on board with all of that, thanks both!

PR #144, opened 2025-07-31
Add Object Instance management endpoints

j616:
# Details
Adds Object Instance management endpoints to support the creation/management of duplicate instances of Objects across multiple storage backends.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5448

# Related PRs
If merged before https://github.com/bbc/tams/pull/115, add the new endpoints to the auth logic listings in that PR. If merged after, add that logic in this PR

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/adr/0038-improved-storage-management.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Decision Outcome
+
+Chosen options:
+
+* Option 1b: Add `get_urls` management to the Objects endpoint
+* Option 2b: Manage additional Object storage via a Objects storage endpoint
+* Option 3b: Call Object Instance management endpoint `instances`
+
+These options have been chosen because they provide clearer boundaries between Media Objects and Segments in the data model and it's implementation.
```
samdbmg: Nit: typo
```suggestion
These options have been chosen because they provide clearer boundaries between Media Objects and Segments in the data model and its implementation.
```

docs/adr/0038-improved-storage-management.md
```diff
@@ -0,0 +1,123 @@
...
+They should avoid confusion arrising from changes to one Flow impacting another.
+And they minimise un-needed changes to the API and common workflows.
+
+### Implementation
+
+{Once the proposal has been implemented, add a link to the relevant PRs here}
```
samdbmg: (Figure this might be on your radar already, but just in case)
```suggestion
Implemented by <https://github.com/bbc/tams/pull/144>
```

docs/adr/0038-improved-storage-management.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Pros and Cons of the Options
+
+### Option 1a: Manage `get_urls` via the Flows endpoints
+
+This option would see us add support for in-place editing of `get_urls` and see the edits propogated to other segments making use of the same Media Object.
```
samdbmg: Nit: typo
```suggestion
This option would see us add support for in-place editing of `get_urls` and see the edits propagated to other segments making use of the same Media Object.
```

docs/adr/0038-improved-storage-management.md
```diff
@@ -0,0 +1,123 @@
...
+
+* Good, because it somewhat matches existing patterns for updating `get_urls`
+* Good, because it would remove a race condition of the delete & re-create pattern with Object garbage-collection in implementations
```
samdbmg, line 58: Not entirely sure I follow this one? I think we've talked about this before, but I can't see how this avoids that, and a dedicated endpoint doesn't? Surely tying the URL management to the object avoids that better, because (presumably) you can conduct operations on the object even if it's ref count is now zero?

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+{
+  "limit": 1,
+  "storage_id": "60af2ab4-e8a5-4c65-a09b-d35983680315"
+}
+```
+
+Example response:
+
+```json
+{
```
samdbmg, line 32: Should the `storage_id` appear in here too, if it isn't mandatory when making the request?

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+      "put_url": {
+        "url": "https://example.store.com/tams-e2b89b02-21e7-5f9d-aa2d-db38b01453c9/846023d3-612d-5014-bc47-88f6eb2d04bb?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=0&X-Amz-Date=20230316T120329Z&X-Amz-Expires=300&X-Amz-SignedHeaders=content-type%3Bhost&X-Amz-Signature=0",
+        "content-type": "video/mp2t"
+      }
+    }
+  ]
+}
+```
+
+In the response above, you can see a "pre" action alongside the the Media Object.
```
samdbmg: At some point we should consider whether this is a behaviour anyone is using, and if not, whether it's a keeper. Not here though!

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+As the Segment it was used in in `flowId` started at `20:0`, but in `flowId2` it is placed at `165:0`, we must set a `ts_offset` of `145:0`.
+For more information on `ts_offset`, see [here](https://bbc.github.io/tams/7.0/index.html#/operations/GET_flows-flowId-segments).
+
+### Duplicating an Existing Object (Client Managed)
+
+There are many reasons a client may want to create a duplicate instance of a Media Object.
+To create a backup.
+To create copies that are physically or logically closer to other systems.
+To move content to archive storage.
+All while being able to refer to the this collection of duplicates with the same Media Object ID in Flow Segments.
```
samdbmg: Nit: typo
```suggestion
All while being able to refer to this collection of duplicates with the same Media Object ID in Flow Segments.
```

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+
+There are two methods of creating duplicate instances of a Media Object.
+The first requires the client to manage the duplication process.
```
samdbmg: Should we have an ADR point on client vs server vs both?

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+{
+  "storage_id": "323367fd-21bb-4f2e-ad38-faf048c4ccfc"
+}
+```
+
+The available instances of the Media Object will now be advertised in `get_urls` on the `/objects/{objectId}` endpoint and on the `/flows/{flowId}/segments` for all Flow Segments which use the Media Object.
+
+### Duplicating an Existing Object (Server Managed)
+
+Some TAMS implementations may support a second, server managed, method of duplicating Media Objects.
```
samdbmg: How do we signal whether or not this is available?

I'm also unsure how we strike a balance between providing this sort of useful functionality against making the store too clever and therefore too difficult to implement (especially as an interface - e.g. at the edge of a MAM or PAM) against providing too many permutations that block interoperability. Would suggesting that the store may choose to do some duplicating itself (e.g. if your MAM has some replication rules) but there isn't an API call for it?

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+
+```json
+{
+  "storage_id": "323367fd-21bb-4f2e-ad38-faf048c4ccfc"
+}
+```
+
+### Deleting an Object Instance
+
+Specific instances of a Media Object can be deleted by a DELETE request to `/objects/{objectId}/get_urls` with the relevant `storage_id` in the query string.
```
samdbmg, line 111: What happens if you delete the last storage location?

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Deployment Considerations
+
+### Security
+
+The approach to supporting multiple Media Object instances in TAMS enables efficient re-use of media, changing ownership through the lifecycle of the media, and self-service re-location of media to meet the purposes of individual users
+With this increased flexibility comes the potential for new attack vectors for malicious actors.
+
+Consider a Flow A with its Media Objects.
+A malicious actor has read access to Flow A, but not read access.
```
samdbmg: Nit: I'm guessing this is what it should say?
```suggestion
A malicious actor has read access to Flow A, but not write access.
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1785,7 +1785,34 @@ paths:
...
+        - The `sample_offset` SHOULD be zero.
+
+        A TAMS SHOULD reject registrations of Flow Segments with a 400 error response if it references a newly created media object in the local TAMS storage that was not intended to be used for the Flow. A TAMS SHOULD accept Flow Segments that reference an existing media object in the local TAMS storage that was originally created for another Flow.
+
+        A TAMS MAY support media objects that are held in external storage in another TAMS or other media storage system. The Flow Segment may in that case require the `get_urls` property to provide the information needed by clients to access the media object.
+
+        Clients MAY modify Flow Segments, but this should only be done in exceptional circumstances to correct metadata as such operations will likely break the idempotency of Segments. Properties of Media Objects, such as `get_urls`, SHOULD be modified via the `/objects` endpoints. They SHOULD NOT be modified at this endpoint, and TAMS instences SHOULD reject such requests with a `400` error.
```
samdbmg: Nit: typo
```suggestion
        Clients MAY modify Flow Segments, but this should only be done in exceptional circumstances to correct metadata as such operations will likely break the idempotency of Segments. Properties of Media Objects, such as `get_urls`, SHOULD be modified via the `/objects` endpoints. They SHOULD NOT be modified at this endpoint, and TAMS instances SHOULD reject such requests with a `400` error.
```

api/schemas/object-storage-post.json
```diff
@@ -0,0 +1,13 @@
+{
+    "title": "Flow Storage Post",
+    "description": "Post data for the flow storage endpoint",
```
samdbmg: Nit: Copy-paste
```suggestion
    "title": "Object Storage Post",
    "description": "Post data for the storage endpoint for working with existing objects",
```

docs/adr/0038-improved-storage-management.md
j616: If we allow `get_urls` to be edited in place, which we don't at the moment, you don't have to delete the segment and re-create it with the correct metadata. Such a delete & re-create has a race condition where the garbage collector could clean up the un-referenced object before the updated segment has been created.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: We could do. But it felt like wasteful extra data. The client can find the default storage backend, which is used if none is specified, at the `/service/storage-backends` endpoint.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: The following point is in the API spec:
> API instances SHOULD prevent clients from deleting all Object instances. Where clients wish to remove all copies of an Object from the store, they should do so by deleting all Flows or Flow Segments which reference the Object.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: I think there are benefits to client-triggered server-managed duplication (see the security considerations). While I agree it goes beyond service functionality we've previously specified, I think it's often a single request for many SaS object stores. Which would make it less overhead that generating multiple pre-signed urls.

I'm undecided on the best way to signal the capability, though. I think the obvious option would be a flag on the `/service` endpoint, like we have for the event stream mechanisms. Though I agree that multiple permutations could cause issues.

docs/adr/0038-improved-storage-management.md
```diff
@@ -0,0 +1,123 @@
...
+The TAMS specification has always had the ability to advertise multiple URLs for retrieving Media Objects.
+But, so far, there has not been direct support for creating and managing duplicates of Media Objects under the control of a TAMS instance.
+
+An Objects endpoint was added in [ADR0027](./0027-add-objects-api-endpoint.md) that advertises the Flows where a given Object is referenced.
+This has begun a transition from thinking about Objects being heavily tied to a Segment.
+And a move from thinking about "Segment reuse" to "Object reuse".
+Until now, the TAMS specifcation has been unclear on the ownership of `get_urls` being the Segment or Object.
+In particular, whether re-use of an Object should result in `get_urls` and changes to them being reflected across all segments using of them.
+
+This ADR proposes explicitely linking ownership of `get_urls` to Objects, and providing mechanisms for adding and removing controlled and uncontrolled instances of Objects to their `get_urls` list.
```
GeorginaShippey:
```suggestion
This ADR proposes explicitly linking ownership of `get_urls` to Objects, and providing mechanisms for adding and removing controlled and uncontrolled instances of Objects to their `get_urls` list.
```

docs/appnotes/0018-managing-multiple-object-instances.md
j616: Following a meeting on this, we've decided server-managed is likely less complicated to implement overall and mitigates potential security holes. I'll add an ADR capturing these discussions and update the implementation.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: Following a meeting on this, we've decided server-managed is likely less complicated to implement overall and mitigates potential security holes. I'll add an ADR capturing these discussions and update the implementation.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: Following a discussion, we have decided not to add `storage_id` to requests. We haven't identified any convincing use case that aren't covered by the `/service/storage-backends` endpoint and that justify the additional data in responses to storage allocation requests.

api/TimeAddressableMediaStore.yaml
```diff
@@ -2009,6 +2122,153 @@ paths:
...
+    delete:
+      summary: Delete a Media Object instance
+      description: |
+        Delete an instance of a media object.
+
+        One of `storage_id` or `label` MUST be specified in the query parameters. `storage_id` SHOULD be used where `controlled` is `True` for the instance.
+
+        API instances should remove the Media Object instance from the `get_urls` list and then, if the instance is controlled, delete the object instance from storage.
+
+        API instances SHOULD prevent clients from deleting all Object instances. Where clients wish to remove all copies of an Object from the store, they should do so by deleting all Flows or Flow Segments which reference the Object.
```
GeorginaShippey: If we end up in a situation where the only available instance is uncontrolled it's possible for the rug to get pulled out from under you. Perhaps it needs to be stipulated that there must be at least one controlled instance available.

api/TimeAddressableMediaStore.yaml
```diff
@@ -2009,6 +2122,153 @@ paths:
...
+      tags:
+        - Objects
+      parameters:
+        - name: objectId
+          in: path
+          description: The Media Object identifier.
+          required: true
+          schema:
+            type: string
+        - name: storge_id
```
GeorginaShippey:
```suggestion
        - name: storage_id
```

docs/appnotes/0018-managing-multiple-object-instances.md
```diff
@@ -0,0 +1,221 @@
...
+      "bucket_id": "tams-e2b89b02-21e7-5f9d-aa2d-db38b01453c9",
+      "put_url": {
+        "url": "https://example.store.com/tams-e2b89b02-21e7-5f9d-aa2d-db38b01453c9?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=0&X-Amz-Date=20230316T120329Z&X-Amz-Expires=300&X-Amz-SignedHeaders=host&X-Amz-Signature=0",
+        "body": "<CreateBucketConfiguration xmlns=\"http://s3.amazonaws.com/doc/2006-03-01/\">\n    <LocationConstraint>default</LocationConstraint>\n</CreateBucketConfiguration >\n"
+      }
+    }
+  ],
+  "media_objects": [
+    {
+      "object_id": "tams-e2b89b02-21e7-5f9d-aa2d-db38b01453c9/846023d3-612d-5014-bc47-88f6eb2d04bb",
```
GeorginaShippey, line 35: I can see a few cases where the object_id contains a bucket as a prefix - I don't think this makes sense anymore considering multiple object instances in different locations share the object_id

api/TimeAddressableMediaStore.yaml
j616: As discussed. We will still have to handle cases where the first registered instance is uncontrolled. I'll add a note that stores may also want to make sure at least one controlled instance remains, where possible.

docs/appnotes/0018-managing-multiple-object-instances.md
j616: Unfortunately, the bucket name is used to identify if the `pre` action in a `/flows/{flowID}/storage` request needs to be run to create a bucket for a given object on initial upload of content. Removing this would require a breaking change to that part of the API, which I'd like to avoid for now.

samdbmg (approved):
Rework LGTM

PR #146, opened 2025-08-04
Clarify behaviour on conflicting Flow/Source metadata

j616:
# Details
Add note that store implementations should verify that creation/modification of Flows will not conflict with Sources

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5456

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -924,7 +924,10 @@ paths:
           description: Bad request. Invalid query options.
     put:
       summary: Create or Replace Flow
-      description: Create or replace the flow metadata.
+      description: |
+        Create or replace the flow metadata.
+
+        Store implementations SHOULD verify that Flow metadata is compatible with the associated Source. Store implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict. Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
       operationId: PUT_flows-flowId
       tags:
         - Flows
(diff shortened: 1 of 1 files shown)


api/TimeAddressableMediaStore.yaml
```diff
@@ -924,7 +924,10 @@ paths:
           description: Bad request. Invalid query options.
     put:
       summary: Create or Replace Flow
-      description: Create or replace the flow metadata.
+      description: |
+        Create or replace the flow metadata.
+
+        Store implementations SHOULD verify that Flow metadata is compatible with the associated Source. Store implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict. Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
```
samdbmg: Should we offer an example of what would or wouldn't be allowed? Or a cross-link to the Flow/Source model or MS-04 (although maybe that's not entirely helpful?)

I guess that's SHOULD rather than MUST to leave space for a _really really_ good reason to allow the change?

samdbmg: > I guess that's SHOULD rather than MUST to leave space for a really really good reason to allow the change?

I had this one backwards: conflicting changes aren't allowed at all - the SHOULD means an API should implement the check but doesn't have to! (thanks @j616 for the nudge in the right direction)

PR #149, opened 2025-09-08
Audit and update TAMS documentation and specification

j616:
# Details
This PR audits the TAMS repository for consistency, clarity, and correctness. While it has made descriptions more explicit, it has not made any practical changes to the API. This PR has been broken into clear, specific commits where possible.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5391

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,13 +132,13 @@ Notice that the `timerange` still refers to the Flow timeline (and `0:50...` etc
 In this way a simple copy-on-write mechanic can be applied to the store, for example when taking "clips" from multiple Flows and assembling them, the original Media Objects (and therefore essence) can be referenced, and new Media Objects are only required to handle changes, for example transitions.
 In the diagram below, portions of Flow X and Flow Y are combined to form Flow Z, along with some rendered transitions which are "new" media, and new Media Objects accordingly.
 
-![Graphic showing the Flow timeline and Flow Segments of Flows X, Y and Z, where Z is composed of a mix of re-used segments and new media](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ.drawio.png)
+![Graphic showing the Flow timeline and Flow Segments of Flows X, Y and Z, where Z is composed of a mix of re-used Segments and new media](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ.drawio.png)
 
-In practice a client may need to read (and potentially decode, in the case of inter-frame video codecs) the entire Object, discarding the unwanted grains and returning those selected by the segment to the consumer.
-This can be handled by mapping the Object's timeline into the Flow timeline by adding `ts_offset` to each timestamp inside the Media Object, and then discarding grains when the resulting timestamp falls outside the given `timerange` for the segment.
-Alternatively if `sample_offset` and `sample_count` are set on the segment, a client may count the samples read from the Object and discard accordingly, although care must be taken with codecs that use temporal re-ordering, since the `sample_offset` and `sample_count` refer to the presentation timeline, rather than the decode timeline.
+In practice a client may need to read (and potentially decode, in the case of inter-frame video codecs) the entire Object, discarding the unwanted grains and returning those selected by the Segment to the consumer.
+This can be handled by mapping the Object's timeline into the Flow timeline by adding `ts_offset` to each timestamp inside the Media Object, and then discarding grains when the resulting timestamp falls outside the given `timerange` for the Segment.
+Alternatively if `sample_offset` and `sample_count` are set on the Segment, a client may count the samples read from the Object and discard accordingly, although care must be taken with codecs that use temporal re-ordering, since the `sample_offset` and `sample_count` refer to the presentation timeline, rather than the decode timeline.
 
-![Graphic showing the Media Objects making up Flow Z (above) and their segments, along with the selected grains](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ-subsegments.drawio.png)
+![Graphic showing the Media Objects making up Flow Z (above) and their Segments, along with the selected grains](./docs/images/Flow%20and%20Media%20Timelines-Flow%20XYZ-subsegments.drawio.png)
 
 The diagram above shows three of the Media Objects used by Flow Z and a smaller portion of the Flow Z timeline considered previously, with both the Media Object timeline (and `media_ts`) and Flow timeline (`segment_ts`) of each grain in the Media Object along the bottom.
 The `timerange`/`ts_offset` approach is illustrated by the following pseudocode, which shows how a client might apply the `ts_offset` to each `media_ts`, then validate whether it is inside the given `timerange`.
@@ -155,7 +155,7 @@ else:
                                      # (and `segment_ts = 3:500000000 or 3.5sec)`
 ```
 
-The `sample_offset` and `sample_count` approach can be applied using FFmpeg's `-ss` and `-t` options (after converting them into durations), and the resulting output timeline produced by setting `-output_ts_offset` to the start of the segment.
+The `sample_offset` and `sample_count` approach can be applied using FFmpeg's `-ss` and `-t` options (after converting them into durations), and the resulting output timeline produced by setting `-output_ts_offset` to the start of the Segment.
 Note however that for this to be frame-accurate in codecs using temporal re-ordering, the input must be transcoded.
 For example for Media Object Y05 in Flow Z above:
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -53,6 +53,10 @@ But if that is not needed, media work can be done purely by reference.
 Grains are grouped into Flow Segments, containing for example one second of content, wrapped in a container format such as MPEG-TS.
 The store provides a mechanism to upload and register new Flow Segments, and an interface to request all the Flow Segments covering a particular timerange and their download URLs; an approach inspired by chunked streaming protocols like HTTP Live Streaming.
 
+> [!NOTE]
+> TAMS is not opinionated about what container format is used.
```
samdbmg: Nit: Precision - when you pick one we start having opinions
```suggestion
> TAMS does not specify which container format should be used.
```

api/examples/flow-get-200-image-jpeg.json
```diff
@@ -1,26 +1,29 @@
...
+    "segments_updated": "2023-09-14T09:45:26Z",
+    "description": "Big Buck Bunny",
+    "label": "bbb",
+    "format": "urn:x-tam:format:image",
+    "created_by": "tams-dev",
+    "updated_by": "tams-dev",
+    "codec": "image/jpeg",
+    "container": "image/jpeg",
+    "essence_parameters":
+    {
```
samdbmg, line 16: This isn't a form I've seen before - where does this come from? More generally, is there a lint rule we could build in CI?

api/TimeAddressableMediaStore.yaml
```diff
@@ -1651,15 +1651,17 @@ paths:
...
 
         The flow segment may include timing adjustment information that the client needs to apply when extracting the samples from the media object.
         It may also use a subset of the samples in the media object.
         - The timestamp of a sample on the flow segment's timeline (`segment_ts`) is the timestamp of that sample embedded in or derived from the internal timing of the media object (`media_object_ts`) adjusted by `ts_offset`: `segment_ts = media_object_ts + ts_offset`.
         - In addition, the segment's `timerange` start MUST be at `sample_offset` samples from the start of the presentation timeline of the object, and `timerange` end MUST be at `sample_offset + sample_count` samples.
 
-        Use the pagination options to limit the results to a timerange and/or count.
+        Clients should use the pagination options to limit the results to a timerange and/or count.
+        Service implementations may also limit the results returned.
+        This will be signaled via the paging headers in the response.
```
samdbmg: Nit: typo (I think)
```suggestion
        This will be signalled via the paging headers in the response.
```

api/TimeAddressableMediaStore.yaml
```diff
@@ -1000,7 +1000,7 @@ paths:
               example:
                 $ref: examples/deletion-request-get-200.json
         "204":
-          description: No content. The flow has been deleted and the flow segments have been or will be deleted.
+          description: No content. The flow has been deleted and the flow segments have been or will be deleted. Media objects referenced in other flows will not be deleted. Media objects that are no longer referenced by any segments will be deleted.
```
samdbmg: Do we describe this behaviour anywhere else? It should be part of the contract (at least, the first sentence is obvious, I'm less sure whether we explicitly say the second should also apply)

api/TimeAddressableMediaStore.yaml
```diff
@@ -1414,7 +1414,11 @@ paths:
           $ref: '#/components/responses/trait_resource_info_head_404'
     get:
       summary: Flow Maximum Bit Rate
-      description: Returns the flow max bit rate property.
+      description: |
+        Returns the flow max bit rate property.
+
+        The maximum bit rate of the flow segments in 1000 bits/second.
```
samdbmg: I see this is the same form of words as in `flow-core.json` - I suppose we're trying to say that it's measured in kilobits per second, without having to figure out whether that's 1000 bits or 1024 bits

api/TimeAddressableMediaStore.yaml
```diff
@@ -18,14 +18,14 @@ info:
   title: Time-addressable Media Store
   description: |
     The Time-addressable Media Store (TAMS) is used for storing segmented media flows.
-    The TAMS consists of a media store for the media Flow Segment Objects and a service providing a database index of the Flow Segments.
+    The TAMS consists of one or more Storage Backend for the media Flow Segment Objects and a service providing a database index of the Flow Segments.
```
samdbmg: Nit: typo
```suggestion
    The TAMS consists of one or more Storage Backends for the media Flow Segment Objects and a service providing a database index of the Flow Segments.
```

docs/README.md
```diff
@@ -1,5 +1,29 @@
 # Supporting Documentation
 
+## Application Notes
+
+Application notes are informatative documents describing the recommended usage of the API.
```
samdbmg: Nit: typo
```suggestion
Application notes are informative documents describing the recommended usage of the API.
```

samdbmg (approved):
Overall LGTM - couple of nits and notes inline

Also a few specifics I picked up on the way through:
42ebd51 - `flow-segment-post.json` still contains an object ID in the old style
1e6e363 - not sure I fully understand the logic as to which endpoints do or don't get 403s? E.g. `PUT /flows/{flowId}/read_only` doesn't? Listing endpoints not having it, get endpoints not having it and HEAD not having don't
6a8722b - Still a few references to "flow" (lowercase) - especially the toplevel readme

api/examples/flow-get-200-image-jpeg.json
j616: Its whatever the "format JSON" command in my editor is doing :sweat_smile: I was more worried about the consistent indenting etc tbh

api/schemas/flow-segment-post.json
```diff
@@ -1,59 +1,74 @@
...
-      "type": "array",
-      "items": {
-        "type": "object",
-        "required": [
-          "url"
-        ],
-        "properties": {
-          "url": {
-            "description": "A URL to which a GET request can be made to directly retrieve the contents of the segment. Clients should include credentials if the provide URL is on the same origin as the API endpoint",
+    "title": "Flow Segment",
```
GeorginaShippey: NIT: It would be better for the render if the title also made reference to the fact that this schema is specifically used for POST requests, if not flow-segment-post.json and flow-segment.json both appear with the same title in the UI, in the same vein if the title for flow-segment.json was updated to include the word GET I wouldn't be mad either! There is something similar going on for Webhooks but since that's getting updated elsewhere maybe its best to leave to the other PR.

api/examples/flow-get-200-video-h264.json
```diff
@@ -1,51 +1,58 @@
...
+    "segments_updated": "2023-09-14T09:45:26Z",
+    "description": "Big Buck Bunny",
+    "label": "bbb",
+    "format": "urn:x-nmos:format:video",
+    "created_by": "tams-dev",
+    "updated_by": "tams-dev",
+    "tags":
+    {
+        "input_quality": "contribution",
+        "_tams_segmentation_rate": "10"
```
GeorginaShippey: Sorry to potentially be annoying quite late in the day... _tams_segmentation_rate. Shouldn't this be deprecated as a result of [ADR 22](https://github.com/bbc/tams/blob/941967e5516a5e0c27405a32c6a0a0662b98afc4/docs/adr/0022-flow-bit-rate-properties.md), I see [PR 86](https://github.com/bbc/tams/pull/86/files) implemented segment_duration but did not deprecate the tag in the commonly used tags list in app note 003 nor update any of the examples. Or have I misunderstood something there? Might be appropriate to put into its own PR tbh.

j616: I've just dropped these tags from the examples

api/TimeAddressableMediaStore.yaml
j616: Yep. I think this has been in the spec in some form since the rate parameters were added. As you suggest, I think it's because correct use of "kilo" and "kibi" is somewhat varied in our field!

api/examples/flow-get-200-image-jpeg.json
j616: We should probably add a linter for the json files, though

PR #150, opened 2025-09-19
add status and error fields to webhook records

danjhd:
# Details
A very simple PR to suggest a way to add status and error to the webhooks schema.

# Jira Issue (if relevant)
Jira URL: n/a

# Related PRs
https://github.com/bbc/tams/pull/142
I am targeting this PR into the branch being used for that PR. Do not merge into any other branch.

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/schemas/webhook.json
+++ b/api/schemas/webhook.json
@@ -86,6 +86,14 @@
         "verbose_storage": {
             "description": "Whether to include storage metadata in the `get_urls` property in `flows/segments_added` events. This option is the same as the `verbose_storage` query parameter for the /flows/{flowId}/segments API endpoint.",
             "type": "boolean"
+        },
+        "status": {
+            "description": "Status of the Webhook. A disabled webhook will not POST events.",
+            "type": "string",
+            "enum": [
+                "enabled",
+                "disabled"
+            ]
         }
     }
 }
--- a/api/schemas/webhook.json
+++ b/api/schemas/webhook.json
@@ -90,6 +90,7 @@
         "status": {
             "description": "Status of the Webhook. A disabled webhook will not POST events.",
             "type": "string",
+            "default": "enabled",
             "enum": [
                 "enabled",
                 "disabled"
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -214,7 +214,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "schemas/webhook-with-id.json"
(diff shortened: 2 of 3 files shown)


api/schemas/webhook.json
```diff
@@ -86,6 +86,15 @@
         "verbose_storage": {
             "description": "Whether to include storage metadata in the `get_urls` property in `flows/segments_added` events. This option is the same as the `verbose_storage` query parameter for the /flows/{flowId}/segments API endpoint.",
             "type": "boolean"
+        },
+        "status": {
```
j616, line 90: Is it intentional that this is setable by the client? I can see use cases for that.

api/schemas/webhook.json
```diff
@@ -86,6 +86,15 @@
...
+            "description": "Status of the Webhook. A disabled webhook will not POST events.",
+            "type": "string",
+            "default": "enabled",
+            "enum": [
```
j616, line 94: Should this have `"error"` as an option to follow the pattern of [Flow Delete Requests](https://bbc.github.io/tams/main/index.html#/operations/GET_flow-delete-requests-request-id)? I think the intention of that pattern is that the `error` object is only present when `status` is set to `error`. If we want to follow that pattern, and we want to make this user setable, we might want to think about the best way to indicate the values a client may set and the values a service instance may set. We should also probably describe what a client should do if a webhook enters the `error` status. e.g. could they set it back to `enabled` to try again if they fix an issue on the receiving end? Would there be any benefit to having the separate `created` and `started` states of the Flow Delete Requests here too? Or is it fair to assume the startup time is negligible?

danjhd: I did not think about it to this depth, apologies. You have valid points, my response is "yes"! The difference between `created` and `started` might be a way of indicating that at least one POST has occurred since creation? Just thinking out loud. I am certainly not too opinionated on the enum values. I just thought creating this PR was better than just leaving you to iron out the details ;)

api/schemas/webhook.json
danjhd: Yes, i was intending this to be settable by the client. Not sure how i should reflect this in the spec? I thought by adding it to the generic "webhook" schema that was implied but i should proabbly add something somewhere to be explicit.

j616: I think what you've done is enough. Just wanted to double check as it went beyond reporting errors :)

PR #152, opened 2025-09-29
appnote: Fix error in deleting object instances

samdbmg:
# Details
- Fixes a typo in AppNote 0018 which gave an incorrect path for deleting object instances. Also adds a note about deleting the last instance of an object.
- Fixes a confusion on modifying Flow Segments: modifying the list of `get_urls` is handled by `POST /objects/{objectId}/instances` - modifying `key_frame_count` requires deleting and recreating the segment, and should only be performed exceptionally.

# Jira Issue (if relevant)
Jira URL: N/A (bug report on Slack)

# Related PRs
Fixes errors introduced in #144 (but as-yet unreleased)

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -2015,11 +2015,13 @@ paths:
 
         Clients are expected to decide how to break content into Media Objects, however those Objects SHOULD be large enough to avoid excessive round trip overheads in the underlying store (_e.g._ of the order of several megabytes) and where codecs with temporal re-ordering are used, Object SHOULD contain complete GOPs or decodable units.
 
-        For Media Objects that have been re-used from other Flows, the `sample_offset` and `sample_count` MAY be used to specify part of the Media Object to use:
+        For Media Objects that have been re-used from elsewhere, the `sample_offset` and `sample_count` MAY be used to specify part of the Media Object to use:
           - The `timerange` field indicates the new Segment's position in the Flow
           - The sample associated with the start of the `timerange` MUST be the one selected by `sample_offset`
           - Likewise, the sample associated with the end of the `timerange` MUST be the one selected by `sample_offset + sample_count - 1`, unless `timerange` has an exclusive end, in which case it will be  `sample_offset + sample_count`.
           - The timerange of the Segment MUST NOT overlap any other Segment in the same Flow.
+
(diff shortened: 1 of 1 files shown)


danjhd (approved):
LGTM

api/TimeAddressableMediaStore.yaml
```diff
@@ -1951,11 +1951,10 @@ paths:
...
-        Properties of Media Objects, such as `get_urls`, SHOULD be modified via the [`/objects`](#/operations/GET_objects) endpoints.
-        They SHOULD NOT be modified at this endpoint, and TAMS instences SHOULD reject such requests with a `400` error.
+        The list of instances of an object (and associated `get_urls` entries) can be modified via the [`/objects`](#/operations/GET_objects) endpoints, which provides a mechanism to register new instances of an object.
 
-        If a client needs to modify a Flow Segment, e.g. to correct metadata such as the `key_frame_count` or add additional URLs to `get_urls`, then the client SHOULD first delete the existing Segment and then write a new one.
+        Clients MAY modify Flow Segments, but this should only be done in exceptional circumstances to correct metadata such as `key_frame_count`, as such operations will likely break the idempotency of Segments.
+        If a client needs to modify a Flow Segment, then the client SHOULD first delete the existing Segment and then write a new one.
```
j616, line 2009: Should we also add a note, like those we have elsewhere, warning that setting `key_frame_count` to a value that conflicts with the Object SHOULD be rejected if that object is in use elsewhere? i.e. don't bring references into conflict with each other.

j616 (approved):
One Q. Otherwise LGTM

api/TimeAddressableMediaStore.yaml
samdbmg: Should clients ignore it if the object is re-used? That would seem like the easiest behaviour to implement I suppose

j616: I was thinking something like the last para of the main description on https://bbc.github.io/tams/main/index.html#/operations/PUT_flows-flowId . i.e. allow modification if it won't bring anything into conflict. Reject if it will.

samdbmg: After discussion, tweaked in 0338ef6 to make expectation clearer

PR #157, opened 2025-11-14
Describe mechanism for automated retention management

j616:
# Details
This PR includes an ADR that presents multiple options for the signalling and implementation of retention management, alongside the relevant tags and an Application Note for the chosen options. This PR also adds the tag currently used by the AWS store implementation to the listing.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5483

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/adr/0043-signalling-retention-time.md
```diff
@@ -0,0 +1,246 @@
...
+
+This option would see the signalling/configuration of the feature be via a well known Flow tag.
+
+* Good, because it avoids a spec change
+* Good, because it avoids adding a feature to the spec that would likely need to have optional support
+* Good, because it avoids fragmented compatibility around a common function
+* Good, because it does not make an assumption about where the feature would be implemented (i.e. in the service, or elsewhere)
+* Neutral, because it avoids a proliferation of features in the core spec
+* Neutral, because it results in a common feature being signalled via tags
+* Bad, because it requires support for the feature to be verified on deployment of clients which use it
```
samdbmg, line 134: I think there's a little more nuance in this. Ultimately a content owner has to have rules about the lifecycle of their content (naively, that rule could be "keep all of it forever"), and they own the risks of getting that wrong: either storage cost increase, or loss of irreplaceable material. Given the impact of that last risk, they're going to be pretty cautious about which instructions they trust from users/tools/etc.
Conversely I think the client doesn't really care: it's giving a hint as to how it feels about keeping/discarding the content that may or may not be honoured. tl;dr - I don't think the client depends on this

docs/adr/0043-signalling-retention-time.md
```diff
@@ -0,0 +1,246 @@
...
+
+This option would see the time for a Flow's automated deletion to be signalled with an absolute time.
+
+* Good, because it is immediately obvious when a Flow will be deleted
+* Neutral, because enacting this requires no further calculations
+* Bad, because this mode of operation cannot be statically templated
+
+### Option 6b: Describe a method of deleting an entire Flow at a time offset from its creation time
+
+This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's Creation time.
```
samdbmg: Nit: Using field names to be consistent with the nit below, and make more obvious it's a field
```suggestion
### Option 6b: Describe a method of deleting an entire Flow at a time offset from its `created` time

This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's `created` time.
```

docs/adr/0043-signalling-retention-time.md
```diff
@@ -0,0 +1,246 @@
...
+
+* Good, because it supports static templating (e.g. via Flow Profiles)
+* Neutral, because enacting this requires a simple calculation to identify the actual time for deletion
+* Neutral, because it allows Flows to be assigned an explicit life time
+* Bad, because it doesn't take into account whether the Flow is still being written to
+
+### Option 6c: Describe a method of deleting an entire Flow at a time offset from its Segments updated time
+
+This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's Segments Updated time.
```
samdbmg: Nit: Consistent styling, makes clearer it's a Flow field
```suggestion
### Option 6c: Describe a method of deleting an entire Flow at a time offset from its `segments_updated` time

This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's `segments_updated` time.
```

docs/adr/0043-signalling-retention-time.md
```diff
@@ -0,0 +1,246 @@
...
+
+* Good, because it supports static templating (e.g. via Flow Profiles)
+* Good, because it allows Flows to be assigned an explicit life time beyond when Segments are written to it
+* Neutral, because enacting this requires a simple calculation to identify the actual time for deletion
+
+### Option 6d: Describe a method of deleting an entire Flow at a time offset from the most recent of Segments updated or Metadata updated time
+
+This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's the later of Segments Updated or Metadata Updated time.
```
samdbmg: Nit: Consistent styling, makes clearer it's a Flow field
```suggestion
### Option 6d: Describe a method of deleting an entire Flow at a time offset from the most recent of `segments_updated` or `metadata_updated` time

This option would see the time for a Flow's automated deletion to be signalled with an offset to be applied to the Flow's the later of `segments_updated` or `metadata_updated` time.
```

docs/adr/0043-signalling-retention-time.md
j616: I'm not sure I agree with "I don't think the client depends on this". As you mentioned, there is the consideration of cost management. But also other considerations such as rights expiration, the risk of storing sensitive content longer than required, or the proliferation of content to be included in listings. While I agree that this is strictly speaking the concern of the content owner, I suspect that the writing client will often be the best point in the workflow to configure the life-cycle management for that content.

j616: I'm not sure I agree with "I don't think the client depends on this". As you mentioned, there is the consideration of cost management. But also other considerations such as rights expiration, the risk of storing sensitive content longer than required, or the proliferation of content to be included in listings. While I agree that this is strictly speaking the concern of the content owner, I suspect that the writing client will often be the best point in the workflow to configure the life-cycle management for that content.

docs/adr/0043-signalling-retention-time.md
```diff
@@ -0,0 +1,246 @@
...
+The assumption being that, for more complex use cases, a MAM layer may sit on top of a TAMS service.
+Or a MAM may provide a TAMS API as an interoperable means to interact with the MAM at a basic level.
+Loop recording could be consider a concern for another system, such as a MAM.
+
+To judge this position, we might consider what other types of media management TAMS is currently responsible.
+
+TAMS implementations are currently responsible for the reference counting of Media Objects and deleting the Objects from storage once it is no longer references by any Flow Segments.
+It is vital for the management of Media Objects given that they have no conceptual "owner" within the TAMS system, and thus no other system responsible for their deletion.
+This operation is performed by reference counting Object IDs as opposed to explicit signalling by the client.
+It requires no configuration
```
himslm01: Missing fullstop.

docs/adr/0043-signalling-retention-time.md
samdbmg: After discussion, I think what we have sufficiently conveys that this is a "bad"...but not _that bad_, and it makes sense to source that information from writers even if you don't necessarily use it unmodified (e.g. because your MAM imposes rules).
Let's stick with this.

j616:
I've made a first pass at ADR options for signalling garbage collection and presigned URL timeouts in https://github.com/bbc/tams/pull/157/commits/7fe2a376d5cee385c9b8589b1401af8c5ee263c8 . I've haven't selected any options yet. I personally favour 8a and 9a for the reasons outlined in the document. But I'd appreciate feedback.

j616:
Just an update from our side. We've come to the realisation that this PR is getting very broad in scope and quite unwieldy! We believe the garbage collection and presigned URL timeouts proposals can sensibly be split out into a separate PR/ADR. That would unblock the tag-based retention management aspects which already have consensus and allow this PR to be merged. I'll make that change soon and provide a link to the new PR once available.

PR #167, opened 2026-01-28
Add support for init segments in Flows

### LINKED ISSUE #194: Add support for fMP4 (and init segments in general)
**Description**
While TAMS is agnostic to the container format used by Flow Segments, conventionally most implementations use MPEG-TS. We should explore how to add other common formats, specifically such as fMP4 as used in MPEG-DASH.

However this comes with a snag, that unlike MPEG-TS which is self-initialising, fMP4 requires an "init segment" before playing media. Currently there is nowhere to store that in TAMS, and we should explore options to do so.

**Use Case**
_TODO_

j616:
# Details
Support for init Segments in Flows

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5501

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


api/examples/flow-segments-get-200-init.json
```diff
@@ -0,0 +1,53 @@
...
+    {
+        "object_id": "846023d3-612d-5014-bc47-88f6eb2d04bb",
+        "timerange": "[0:0_10:0)",
+        "get_urls": [
+            {
+                "url": "https://store.example.com/tams-e2b89b02-21e7-5f9d-aa2d-db38b01453c9/846023d3-612d-5014-bc47-88f6eb2d04bb"
+            }
+        ],
+        "init_object": {
+            "id": "9ef06e23-b882-4a93-b005-480de42ef4e9",
```
danjhd: This feels weird to me, and i am happy to be told it is only me but i wanted to flag it just in case. I think this `id` field should be `object_id` as above it not just `id`

api/schemas/flow-segment.json
```diff
@@ -39,11 +39,33 @@
...
+                },
+                "init_object": {
+                    "title": "Initialisation Object",
+                    "description": "The Object containing the initialisation segment required to decode the parent Media Object.",
+                    "unevaluatedProperties": false,
+                    "allOf": [
+                        {
+                            "type": "object",
+                            "required": [
+                                "id"
```
danjhd: As in my comment in the example i think this should be `object_id` not just `id`

api/examples/flow-segments-get-200-init.json
j616: Yeh. I was on the fence with it. Prior convention has been that if the resource type is the same as in the path, we don't prepend it. See `/flows`, `/flow/<id>`, `/source/<id>`, `/webhook/<id>`, etc. The difference here being that were nesting one Object in another. I chose this approach as the hierarchy then reads `init_object.id` rather than `init_object.object_id` which felt a bit redundant. But it's a weakly held opinion.

docs/adr/0045-flow-init-segments.md
```diff
@@ -0,0 +1,196 @@
...
+
+Option 8 would again place restrictions on workflows that TAMS interfaces with, or would require transformation to/from an embedded-segment rendition.
+Such an approach was also identified as being expressly prohibited by some media formats.
+Real-world issues were also identified regarding processing load, and smoothness of playback with some player/decoder implementations where superfluous init segments are provided.
+
+Option 5 provides little/no restrictions on broader workflows.
+It provides good backwards compatibility.
+It provides good feature compatibility with the likes of edit-by-reference.
+It provides minimal burden on client implementations.
+It re-uses existing patterns, keeping the burden on service implementations as low as possible.
```
samdbmg, line 62: I think this is a nice summary of where we got to in this research, and concisely lays out why 5 is the best option

docs/adr/0045-flow-init-segments.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Option 5: Make the init segment an object-level property, which can be uploaded and have a URL generated
+
+As with Option 4, this option would see the init segment managed via the existing Object CRUD mechanisms.
+Rather than associating the init Object with the Flow, this option would see it associated with each Media Object.
+A change in the init segment's Object ID may be used to determine a need to re-provision the init segment in the consuming client.
+
+* Good, because it doesn't require format-specific Flow parameters
+* Good, because it re-uses existing patterns/mechanisms
+* Good, because it allows for different init segments for different parts of the Flow
```
samdbmg, line 135: Worth adding a "Good" about how edit-by-reference "just works" (something only true of options 1, 2, 5 and 8 I think?)

docs/appnotes/0024-using-init-segments.md
```diff
@@ -0,0 +1,56 @@
+# 0024: Using Media with Initialisation Segments in TAMS
+
+This Application Note assumes basic knowledge of how to write/read media to/from a TAMS service.
+
+Some media formats, such as CMAF compatible MPEG-DASH and fMP4, split content into initialisation (init) segments and media segments.
```
samdbmg: Nit: Could you link CMAF to some more info - it's a little more niche than DASH or fMP4

docs/appnotes/0024-using-init-segments.md
```diff
@@ -0,0 +1,56 @@
...
+When the configuration changes, a new init segment may be required.
+While a change in technical parameters would generally require a new Flow in TAMS, some init segment changes may be minor codec parameter changes which would not warrant a new Flow.
+A change in init-segment will normally result in a new "period" in DASH, or a "discontinuity" in HLS.
+Other formats may have similar concepts.
+
+Additionally, technologies implemented as extensions to init segments may result in a new init segment for non-technical changes.
+One example being C2PA signing of live media.
+C2PA uses cryptography to verify the provenance of media.
+One of the two C2PA live profiles places cryptographic keys in the init segment which may be used to verify media segments and their sequencing.
+When that cryptographic key is rotated, a new init segment will be published.
```
samdbmg, line 32: This probably deserves a link to an authoritative source/somewhere to learn more too

docs/appnotes/0024-using-init-segments.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Read
+
+1. Get the Flow Segments listing for the Flow
+2. For each Flow Segment
+    1. If init segment Object ID has changed
+        1. Get init segment Object
+        2. (Re)initialise decoder with init segment
+    2. Get media segment Object
+    3. Play media segment
```
samdbmg: Nit: A reminder that you still have to do this bit might be useful
```suggestion
    3. Perform [media timeline mapping](https://github.com/bbc/tams/#flow-and-media-timelines) as usual in TAMS
    4. Play media segment
```

api/schemas/flow-storage-post.json
```diff
@@ -17,6 +17,10 @@
         "storage_id": {
             "description": "The Storage Backend to allocate storage in. A Storage Backend identifier as advertised at the [/service/storage-backends](#/operations/GET_storage-backends) endpoint. If not set the default, as advertised at the [/service/storage-backends](#/operations/GET_storage-backends) endpoint, will be used if available. An invalid Storage Backend identifier will result in a 400 error.",
             "$ref": "uuid.json"
+        },
+        "content-type": {
+            "description": "The `content-type` to use for the Objects. This parameter MUST only be set where requesting storage for initialisation segments in media formats which require them, and where the mime-type of those initialisation segments differs to that of the media segments. Assumed to be the `container` type of th Flow if not set.",
```
samdbmg: Nit: typo
```suggestion
            "description": "The `content-type` to use for the Objects. This parameter MUST only be set where requesting storage for initialisation segments in media formats which require them, and where the mime-type of those initialisation segments differs to that of the media segments. Assumed to be the `container` type of the Flow if not set.",
```

api/examples/flow-get-200-video-h264-fmp4.json
```diff
@@ -0,0 +1,54 @@
...
+    "label": "bbb",
+    "format": "urn:x-nmos:format:video",
+    "created_by": "tams-dev",
+    "updated_by": "tams-dev",
+    "tags": {
+        "input_quality": "contribution"
+    },
+    "codec": "video/h264",
+    "container": "video/iso.segment",
+    "init_segments": true,
```
samdbmg: This doesn't match the schema as written, which would put it under `essence_parameters`

api/examples/flow-segments-get-200-init.json
samdbmg: I see what you're getting at. Looking at it the other way, making it `object_id` means the `init_object` object keys match with the "general purpose object" keys. Maybe it makes sense as `object_id` since the code that fetches media objects and init segments probably ends up sharing paths, and it saves having a bunch of divergent logging calls etc?

api/schemas/flow-video.json
```diff
@@ -212,6 +212,10 @@
                             "description": "If `true`, the frame rate of the Flow is variable and `frame_rate` MUST NOT be set. If `false` or omitted, the frame rate of the Flow is fixed and `frame_rate` MUST be set.",
                             "type": "boolean",
                             "default": false
+                        },
+                        "init_segments": {
+                            "description": "Whether the Flow makes use of initialisation segments. This parameter MUST be set to `true` if Media Objects have `init_object` populated. If set to `true`, all Media Objects MUST have `init_object` populated. Assume `false` if omitted.",
```
samdbmg, line 217: Thoughts on whether this should be a Flow property? Since it's kinda an artifact of how the `container` and splitting into objects works (and lines up with the Flow GET example in this PR, which seems neater IMO)

docs/appnotes/0024-using-init-segments.md
```diff
@@ -0,0 +1,56 @@
...
+This allows existing Media Object re-use mechanisms (i.e. edit-by-reference) to continue to function.
+Init segments follow the media.
+A client may then detect if the init segment has changed by observing if the init Object ID remains the same or changes between subsequent Flow Segments.
+
+## General Workflows
+
+The general workflows described here refer to init and media segments in isolation from manifests used by specific media formats.
+The approaches have, however, been designed to map to manifests where required.
+
+### Write
```
samdbmg, line 50: Nit: Links or example requests might make this easier to follow - e.g. from reading this it isn't immediately obvious step (2) is `POST /flows/<flowid>/storage?content-type=<init seg MIME type>`

api/schemas/flow-storage-post.json
```diff
@@ -17,6 +17,10 @@
         "storage_id": {
             "description": "The Storage Backend to allocate storage in. A Storage Backend identifier as advertised at the [/service/storage-backends](#/operations/GET_storage-backends) endpoint. If not set the default, as advertised at the [/service/storage-backends](#/operations/GET_storage-backends) endpoint, will be used if available. An invalid Storage Backend identifier will result in a 400 error.",
             "$ref": "uuid.json"
+        },
+        "content-type": {
```
samdbmg: Should this be `content_type` to match the style used elsewhere, rather than matching the HTTP header?

samdbmg (changes requested):
Overall changes make sense and look good. Requesting changes because the examples don't match the schema (although I've inlined a question about the schema).

Also, should we call out around `DELETE /flows/<flowid>/segments` that you should hang on to init segment objects that are referenced by media objects, and delete them when no longer referenced, same as for media objects referenced by Flows?

docs/appnotes/0024-using-init-segments.md
j616: As discussed, I'll add these links in. Though they will point to rendered API docs without the changes until merged

api/schemas/flow-video.json
j616: I've partly placed this here due to how the Profiles proposal has been architected. I think it would need to be at this level to allow the profile to match on this property which has a significant role in compatibility of media.

api/examples/flow-segments-get-200-init.json
danjhd: My main thought was that when you look at the response the object has object_id, get_urls etc... whereas in the same json the structure of the init_object is **`id`**, get_urls etc.. the inconsistancy made it jarring to me. But happy to leave it if this was considered and decided upon deliberately.

j616:
> Also, should we call out around DELETE /flows/<flowid>/segments that you should hang on to init segment objects that are referenced by media objects, and delete them when no longer referenced, same as for media objects referenced by Flows?

@samdbmg I've added a note to here and DELETE /flows/<flowid>. It seems that while we've implied the need for reference counting in the past, we haven't ever called it out against these methods.

api/examples/flow-segments-get-200-init.json
j616: Ah! I'd had the Objects endpoint in my head when writing this. You are right. I'll make the change.

samdbmg (approved):
Rework LGTM following discussion in https://github.com/bbc/tams/issues/194#issuecomment-4691881817

PR #181, opened 2026-04-15
Add AppNote0022 - Architectures for the transforming of media

### LINKED ISSUE #191: Describe architectures for implementing transcode
**Context**
We've done a lot of thinking about the various mechanisms and workflows that could be used for generating proxy Flows, and some proof of concept has been done, but there's no single place where this thinking is documented.

Broadly the approaches we've considered are:
- Generate a proxy as part of an ingest process, and ingest two Flows of the same Source
- Observe the incoming Flow growing (e.g. with webhooks) and trigger a transcode process as segments become available
- Trigger a transcode process for a timerange of the Flow on-demand, as a proxy becomes needed
- Provide a mechanism to transcode and serve (and presumably cache to the store) proxy segments on request _(note this is theoretical - we've never actually tried it)_

Several of these have already been demonstrated in various ways, and the AWS TAMS Tools can do both the (2) transcode-on-webhook and (3) transcode-a-timerange

**Actions**
Write a TAMS application note:
- Outline options for creating proxy Flows as described above
- Document pros/cons/tradeoffs for each option and when you might choose each one - e.g. on-ingest makes sense if you know you'll definitely need the proxy

**Outcome**
Application Note

j616:
# Details
This PR adds an application note describing possible architectures for transforming of media in TAMS. It also includes the creation of a new experimental tag for the triggering of transcodes.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5511

# Related PRs
https://github.com/bbc/tams/pull/182 adds an additional architecture describing transform on request, or Just In Time transcode.

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/appnotes/0022-transcode-architectures.md
+++ b/docs/appnotes/0022-transcode-architectures.md
@@ -1,0 +1,96 @@
+# 0022: Architectures for the transcoding of media
+
+## Abstract
+
+A common use case in TAMS systems is to transcode or transpackage media to provide multiple formats, and qualities.
+This application note discusses the available architectures that may be used when implementing this functionality, and discusses the pros and cons of each.
+
+> [!NOTE]
+> Where this application note refers to "transcode", the transpackage use case is equally applicable unless otherwise stated
+
+## Basic Transcode on Ingest
+
+In this architecture, transcode is the responsibility of the ingest client.
+When media is written to TAMS, the ingest client is configured to transcode the media and write multiple Flows of the same Source to TAMS.
+
+### Pros
+
+* Many ingest clients have added TAMS capability to existing products which already include this capability
+* The ingest client may have access to a higher quality copy of media than the one(s) written to TAMS
+* In many cases, this will be the most simple architecture
+
+### Cons
+
+* Transcode may take place even if likelihood of the transcoded media being used is low
+* Transcode in the ingest client will result in higher ingress bandwidth use compared to transcode within the TAMS Service's compute environment
+
+## Event-triggered Transcode on Ingest
+
+In this architecture, a transcode service subscribes to the webhooks endpoint to receive events for new Segments.
+The transcode service will fetch these Segments, transcode them based on the transcode services configuration, and write them to a new Flow.
+
+### Pros
+
+* Ensures appropriate formats are available for all media, where standard formats are required (e.g. proxies)
+* Requires no action by ingesting or consuming clients
+* Potentially allows for more complex configuration of transcoders
+* Potentially allows for transcode service to be deployed in close (logical) proximity to the TAMS Service
+
+### Cons
+
+* Requires an additional component (the transcode service) in the deployment
+* Transcode may take place even if likelihood of the transcoded media being used is low
+
+## Manually-triggered Transcode
+
+In this architecture, a transcode of an entire Flow is manually triggered by a client.
+This may be done via a call to a 3rd-party API, or by creating a new Flow with the `trigger_transcode` tag set to `immediate`.
+
+### 3rd-party API
+
+In the case of the 3rd-party API approach, a client makes a request to the the transcode service with the Source ID of the media to be transcoded, optionally the timerange to be transcoded, and optionally the output format.
+The transcode client will then register a new Flow of the same Source, and provide the Flow ID to the requesting client.
+
+The transcode client will transcode each Segment in the originating Flow, and write the resultant media back to TAMS against the destination Flow.
+The transcode client should set the destination Flow's `flow_status` tag to `awaiting_content` on initial registration, `ingesting` on first write of a Segment, and `closed_complete` once the transcode is complete.
+
+Where the originating Flow has a `flow_status` of `ingesting`, the transcode client may need to subscribe to `flows/segments_added` webhook events for the originating Flow and transcode new Segments as they become available.
+
+The transcode client may be designed such that all Segments are transcoded in parallel, allowing for faster than real-time transcodes in many cases.
+
+#### Pros
+
+* Keeps control signalling out of the TAMS API
+* Potentially allows for more complex configuration of transcoders
+* Potentially allows for transcode service to be deployed in close (logical) proximity to the TAMS Service
+
+#### Cons
+
+* Requesting clients must integrate with additional APIs
+* More components may mean a more complicated architecture
+
+### Tags
+
+In the case of the Flow tag approach, the transcode service will subscribe to `flows/created` webhook events and identify those with the `trigger_transcode` tag set to `immediate`.
+
+The requesting client will create a new Flow against the existing Source with the required codec and properties, and the `trigger_transcode` tag set to `immediate`.
+The creation of this Flow will result in an event that will be received by the transcode service configured above.
+
+The transcode service will transcode each Segment in the originating Flow, based on the technical properties of the destination Flow, and write the resultant media back to TAMS against the destination Flow.
+The transcode service should set the destination Flow's `flow_status` tag to `awaiting_content` on initial registration, `ingesting` on first write of a Segment, and `closed_complete` once the transcode is complete.
+When the transcode is complete, the `trigger_transcode` tag should be removed from the originating Flow.
+
+Where the originating Flow has a `flow_status` of `ingesting`, the transcode service may need to subscribe to `flows/segments_added` webhook events for the originating Flow and transcode new Segments as they become available.
+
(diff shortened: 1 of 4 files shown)


docs/appnotes/0022-transcode-architectures.md
```diff
@@ -0,0 +1,96 @@
...
+A common use case in TAMS systems is to transcode or transpackage media to provide multiple formats, and qualities.
+This application note discusses the available architectures that may be used when implementing this functionality, and discusses the pros and cons of each.
+
+> [!NOTE]
+> Where this application note refers to "transcode", the transpackage use case is equally applicable unless otherwise stated
+
+## Basic Transcode on Ingest
+
+In this architecture, transcode is the responsibility of the ingest client.
+When media is written to TAMS, the ingest client is configured to transcode the media and write multiple Flows of the same Source to TAMS.
```
samdbmg: Should we mention something like this, as a reminder?
```suggestion
When media is written to TAMS, the ingest client is configured to transcode the media and write multiple Flows of the same Source to TAMS.
Ingest clients must take care that timing is aligned between all generated Flows of the Source.
```

samdbmg (approved):
Nice, LGTM. One suggestion inlined

docs/appnotes/0022-transform-architectures.md
```diff
@@ -0,0 +1,97 @@
...
+Where the originating Flow has a `flow_status` of `ingesting`, the transcode client may need to subscribe to `flows/segments_added` webhook events for the originating Flow and transcode new Segments as they become available.
+
+The transcode client may be designed such that all Segments are transcoded in parallel, allowing for faster than real-time transcodes in many cases.
+
+#### Pros
+
+* Keeps control signalling out of the TAMS API
+* Potentially allows for more complex configuration of transcoders
+* Potentially allows for transcode service to be deployed in close (logical) proximity to the TAMS Service
+
```
GeorginaShippey, line 72: I think the 'optionally the timerange to be transcoded' is being underemphasized/undersold here, this allows for a fine grained approach and reduces the chances of 'Transcode may take place even if likelihood of the transcoded media being used is low'. For a file based / non live workflow we now have the ability to do a partial proxy, rather than the full flow, which for a long running ingest would be a non-insignificant. And for the live edge it allows for a basic on demand transcode approach where the client can use webhooks to subscribe to segments being added to a flow, to then send requests the transcode service for the new timerange, to then wait to be notified again via webhooks once the transcoded segment has become available. The transcode service would have to track multiple transcode requests in order to manage flow_status and trigger_transcode appropriately.

docs/appnotes/0022-transform-architectures.md
```diff
@@ -0,0 +1,97 @@
...
+#### Cons
+
+* Requesting clients must integrate with additional APIs
+* More components may mean a more complicated architecture
+
+### Tags
+
+In the case of the Flow tag approach, the transcode service will subscribe to `flows/created` webhook events and identify those with the `trigger_transcode` tag set to `immediate`.
+
+The requesting client will create a new Flow against the existing Source with the required codec and properties, and the `trigger_transcode` tag set to `immediate`.
```
GeorginaShippey, line 87: Another potential con (if not already covered by the below) is that users may specify a new Flow to be transcoded that is not supported by the transcode service, there would be no neat way to signal this failure in the API currently with the tag values that already exist.

j616: I suppose a `501 Not Implemented` on PUTing the Flow (with `trigger_transcode` tag) with further explanation in the body might be appropriate? But that would only really work for deep integration. So maybe a further tag would be required

GeorginaShippey: I would steer away from the 501 because as you say it requires deep integration which I don't think this option is going for. I think maybe a new tag is the right way to go - and I think this also helps in the case of the 3rd party API too as there could be failure cases as well. Maybe something like transcode_status, then even if trigger_transcode gets scrubbed so the transcode service can be released from processing it, there is still some record of how the flow has come about. Flow_status should then be set to closed_complete... or closed_incomplete??? Depends what we are trying to signal there. Hesitant to make any additions to flow_status for complexity reasons.

docs/appnotes/0022-transform-architectures.md
```diff
@@ -0,0 +1,97 @@
...
+In this architecture, transcode is the responsibility of the ingest client.
+When media is written to TAMS, the ingest client is configured to transcode the media and write multiple Flows of the same Source to TAMS.
+Ingest clients must take care that timing is aligned between all generated Flows of the Source.
+
+### Pros
+
+* Many ingest clients have added TAMS capability to existing products which already include this capability
+* The ingest client may have access to a higher quality copy of media than the one(s) written to TAMS
+* In many cases, this will be the most simple architecture
+
```
johnbilt, line 25: There's another possible pro here which is you can upload the proxy first making it available faster and then upload the high res later.  Good for low bandwidth connections and for getting media to a user as fast as possible to start working with

docs/appnotes/0022-transform-architectures.md
```diff
@@ -0,0 +1,97 @@
...
+### Cons
+
+* Transcode may take place even if likelihood of the transcoded media being used is low
+* Transcode in the ingest client will result in higher ingress bandwidth use compared to transcode within the TAMS Service's compute environment
+
```
johnbilt, line 31: Also requires additional compute time on the ingest client.  Particularly key if running on a user device rather than offloading to another device to run in the background

docs/appnotes/0022-transform-architectures.md
j616: Agreed on not overloading `flow_status`. I'm also concious of overloading `trigger_transcode`. If we accept PR #182 , it will be communicating the type of transcode. We could use a `transcode_status` tag that follows the approach of the [webhook `status` parameter](https://bbc.github.io/tams/8.0/index.html#/operations/GET_webhooks-webhookId). Though we don't currently support dicts in tag values. So we wouldn't be able to have a `transcode_error` counterpart. I'm also quite aware of how many tags would be in play! Another option might just to be a basic `transcode_error` string who's presence signals an error state?

docs/appnotes/0022-transform-architectures.md
j616: Added extra bullet point to capture this

docs/appnotes/0022-transform-architectures.md
j616: Added extra bullet to capture this

docs/appnotes/0022-transform-architectures.md
j616: Added extra bullet to capture this

docs/appnotes/0022-transform-architectures.md
j616: As discussed. The obvious approaches feel unsatisfactory for various reasons. Solving this will need more thought. But we should approach that when users come back to us with clear use cases and requirements. I'll add a "con" item that there's currently no clear way to communicate errors in-band.

GeorginaShippey: Yes, happy with the above!

docs/appnotes/0022-transform-architectures.md
```diff
@@ -0,0 +1,111 @@
...
+## Abstract
+
+A common use case in TAMS systems is to transform media to provide multiple formats, qualities, or modify media in some way.
+This application note discusses the available architectures that may be used when implementing this functionality, and discusses the pros and cons of each.
+
+> [!IMPORTANT]
+> This application note is primarily targeted at transcode, and transpackage operations.
+> The architectures described in this application note may be applied to other transformations, but they may have additional considerations not described here.
+> One important consideration is that if the media changes editorially (e.g. through a colour correction process) then the resultant media is considered a new Source and should be registered in the TAMS Service as such.
```
samdbmg, line 11: Nit: Might be useful to add a link to more guidance here
```suggestion
> One important consideration is that if the media changes editorially (e.g. through a colour correction process) then the resultant media is considered a new Source and should be registered in the TAMS Service as such.
> See [NMOS MS-04](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.1._Basic_Media_Operations.html) for some practical guidance on when a new Source is created.
```

samdbmg (approved):
Changes LGTM, thanks

PR #182, opened 2026-04-15
Add option for Transform on Request to AppNote0022

### LINKED ISSUE #191: Describe architectures for implementing transcode

j616:
# Details
This PR adds an additional transform architecture to AppNote0022, for transcode on request. Or Just In Time transform. As this architecture is more experimental, it has been split into its own PR.

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5511

# Related PRs
Depends on https://github.com/bbc/tams/pull/181

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/docs/appnotes/0022-transform-architectures.md
+++ b/docs/appnotes/0022-transform-architectures.md
@@ -127,6 +127,7 @@ Where the originating Flow has a `flow_status` of `ingesting`, new virtual Segme
 
 ### Pros
 
+* Transcode only takes place when a client attempts to use the transcoded media
 * Requesting clients integrate with fewer APIs
 * Transcode capability to be deployed in close (logical) proximity to the rest of the TAMS Service, due to its deep integration with it
 * Fewer components potentially means a less complicated architecture overall
@@ -136,6 +137,7 @@ Where the originating Flow has a `flow_status` of `ingesting`, new virtual Segme
 * Adds control signalling to the TAMS API
 * Potentially impedes more complex configuration of transcoders
 * Deep integration adds more complexity to TAMS Service implementations
+* Segments need to be short enough and transcode fast enough to respond to Object GET in a timely manner
 
 ### Extension - Segment Duration
 
@@ -143,7 +145,7 @@ While basic implementations may wish to maintain the Segment duration of the ori
 This may be configured by setting the `segment_duration` in the new Flow's metadata.
 The Service should carry out due diligence that the requested Segment duration is valid against restrictions such as possible GOP lengths.
 
-This approach could, for example, allow media to go from single frame per Segment, to long-GOP, and visa-versa.
+This approach could, for example, allow media to go from single frame per Segment, to long-GOP, and vice-versa.
 This capability may be particularly effective in the transpackaging use case, where different size segments may be generated with minimal overheads.
 
 ### Extension - Store Write-Back
--- a/docs/appnotes/0022-transform-architectures.md
+++ b/docs/appnotes/0022-transform-architectures.md
@@ -109,3 +109,51 @@ The body of the response may include further details on the nature of the error.
 * Potentially impedes more complex configuration of transcoders
 * More components may mean a more complicated architecture
 * There are no obvious ways to communicate errors in-band where the transcode service is not deeply integrated that don't have significant drawbacks
+
+## Transcode on Request
+
+This architecture sees transcode of media being triggered just-in-time when a consuming client attempts to read media segments in a format not currently available on-disk.
+
+> [!NOTE]
+> This option likely requires deep integration into the TAMS Service due to the potential creation and removal of Media Object Instances.
+
+The requesting client will create a new Flow against the existing Source with the required codec and properties, and the `trigger_transcode` tag set to `virtual`.
+The TAMS Service will populate Segments and Objects correlating with the timeline range covered by the original Segments.
+These Objects will have instances that are "virtual".
+That is to say that the Objects do not exist on disk.
+When a GET request is performed with the instance URL, the Service will perform the transcode to create the segment and return it, as if the media existed on disk.
+
+Where the originating Flow has a `flow_status` of `ingesting`, new virtual Segments should be created as new originating Segments are registered.
+
+### Pros
+
+* Requesting clients integrate with fewer APIs
+* Transcode capability to be deployed in close (logical) proximity to the rest of the TAMS Service, due to its deep integration with it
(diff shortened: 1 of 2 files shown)


docs/appnotes/0022-transform-architectures.md
```diff
@@ -94,3 +94,51 @@ The transcode service may be designed such that all Segments are transcoded in p
...
+
+## Transcode on Request
+
+This architecture sees transcode of media being triggered just-in-time when a consuming client attempts to read media segments in a format not currently available on-disk.
+
+> [!NOTE]
+> This option likely requires deep integration into the TAMS Service due to the potential creation and removal of Media Object Instances.
+
+The requesting client will create a new Flow against the existing Source with the required codec and properties, and the `trigger_transcode` tag set to `virtual`.
+The TAMS Service will populate Segments and Objects correlating with the timeline range covered by the original Segments.
```
samdbmg, line 121: I'm trying to hint here that you might implement this without creating anything durable: instead rewriting the response of a database query to figure out where the segment boundaries are. But I'm not sure my attempt is any better!
```suggestion
From a client's perspective, Segments and Objects are visible in the new Flow corresponding with the   timerange covered by the original Flow.
```

docs/appnotes/0022-transform-architectures.md
```diff
@@ -94,3 +94,51 @@ The transcode service may be designed such that all Segments are transcoded in p
...
+These Objects will have instances that are "virtual".
+That is to say that the Objects do not exist on disk.
+When a GET request is performed with the instance URL, the Service will perform the transcode to create the segment and return it, as if the media existed on disk.
+
+Where the originating Flow has a `flow_status` of `ingesting`, new virtual Segments should be created as new originating Segments are registered.
+
+### Pros
+
+* Requesting clients integrate with fewer APIs
```
samdbmg, line 135: Small addition to call out the Con this mitigates from above
```suggestion
* Transcode only takes place when a client attempts to use the transcoded media
* Requesting clients integrate with fewer APIs
```

docs/appnotes/0022-transform-architectures.md
```diff
@@ -94,3 +94,51 @@ The transcode service may be designed such that all Segments are transcoded in p
...
+* Transcode capability to be deployed in close (logical) proximity to the rest of the TAMS Service, due to its deep integration with it
+* Fewer components potentially means a less complicated architecture overall
+
+### Cons
+
+* Adds control signalling to the TAMS API
+* Potentially impedes more complex configuration of transcoders
+* Deep integration adds more complexity to TAMS Service implementations
```
samdbmg, line 143:
```suggestion
* Deep integration adds more complexity to TAMS Service implementations
* Segments need to be short enough and transcode fast enough to respond to Object GET in a timely manner
```

docs/appnotes/0022-transcode-architectures.md
```diff
@@ -94,3 +94,51 @@ The transcode service may be designed such that all Segments are transcoded in p
...
+* Potentially impedes more complex configuration of transcoders
+* Deep integration adds more complexity to TAMS Service implementations
+
+### Extension - Segment Duration
+
+While basic implementations may wish to maintain the Segment duration of the originating Flow, a more extensive implementation may allow output Segment duration to be configured.
+This may be configured by setting the `segment_duration` in the new Flow's metadata.
+The Service should carry out due diligence that the requested Segment duration is valid against restrictions such as possible GOP lengths.
+
+This approach could, for example, allow media to go from single frame per Segment, to long-GOP, and visa-versa.
```
samdbmg: Nit: Typo
```suggestion
This approach could, for example, allow media to go from single frame per Segment, to long-GOP, and vice-versa.
```

samdbmg (approved):
Looks good - I've suggested a couple of small wording tweaks

docs/appnotes/0022-transform-architectures.md
j616: I think my concern with this wording is while it captures the apparent behaviour, it doesn't explicitly describe anything actionable. We could add a further sentence along the lines of "How the Service populates these Segments is up to the implementation". But that also feels a touch clunky.

samdbmg: Ok, let's leave this as-is then

docs/appnotes/0022-transform-architectures.md
```diff
@@ -112,7 +112,7 @@ The body of the response may include further details on the nature of the error.
 
 ## Transcode on Request
 
-This architecture sees transcode of media being triggered just-in-time when a consuming client attempts to read media segments in a format not currently available on-disk.
+This architecture sees transcode/transpackage of media being triggered just-in-time when a consuming client attempts to read media segments in a format not currently available on-disk.
```
samdbmg, line 115: Nit: Given our discussion regarding #181, is "transformation" a better word to use here?

samdbmg (approved):
Changes LGTM

docs/appnotes/0022-transform-architectures.md
j616: As per the discussion in the TSC Slack, I've left this as transcode/transpackage as the signalling method only really supports those types of operation. I added the extension to call out that the Virtual Objects approach may be applied to other forms of transform, though. But would require a different approach to configuration.

PR #173, opened 2026-02-09
Describe naming convention for `role` in collections

samdbmg:
# Details
- Describes a naming convention for the `role` field in collections of Flows and Sources
- Adds an ADR discussing how that convention came to be
- Reworks examples throughout to follow the convention

This is a first pass off the back of a discussion about handling broadcast channel ingests: I'd definitely appreciate a wider perspective!

# Jira Issue (if relevant)
Jira URL: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5509

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

- [X] PR completes task/fixes bug
- [ ] API version has been incremented if necessary
- [X] ADR status has been updated, and ADR implementation has been recorded
- [X] Documentation updated (README, etc.)
- [X] PR added to Jira Issue (if relevant)
- [ ] Follow-up stories added to Jira

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] PR completes task/fixes bug
- [ ] Design makes sense, and fits with our current code base
- [ ] Code is easy to follow
- [ ] PR size is sensible
- [ ] Commit history is sensible and tidy

# Info on PRs
The checks above are guidelines. They don't all have to be ticked, but they should all have been considered.

PR #183, opened 2026-04-15
Source collected by query parameter

### LINKED ISSUE #214: Provide a way to list Sources that are collected by a particular Source
**Context**
Currently at the Source level there is no filtering on the `collected_by` field.
When vendors are building systems which list content stored in TAMS then they are only looking for the Sources at the top of the content tree.
In many cases this is the multi-Source, however due to single essence (eg audio only) workflows then this is not a reliable method of finding the top level content.
To achieve this they must go through every result and look to see if the `collected_by` field is missing as this then indicates that it is the top of a Source collection.
This process can also throw any form of sensible pagination in a client application as when requesting content from the TAMS API it is not known how many results will be retained or discarded.

**Actions**
Decide how filtering by `collected_by` should work, and write an ADR and associated spec change
Plan to make the matching change to Flows for consistency and covienience

**Outcome**
- ADR
- New query string parameter

johnbilt:
# Details
Currently at the Source level there is no filtering on the `collected_by` field.
When vendors are building systems which list content stored in TAMS then they are only looking for the Sources at the top of the content tree.
This PR is to add the ability to filter Sources (and Flows) to find the content which is not collected by another Source and therefore is the top level.

# Jira Issue (if relevant)
Jira URL:

# Related PRs

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


docs/adr/0049-source-collected_by-query-parameter.md
```diff
@@ -0,0 +1,102 @@
...
+* Good: Uses an existing field which is already present in the API and provided consistency between different areas
+* Bad: Requires the filtering by ID to be a logical query parameter
+
+### Flow level implications
+
+At the Flow level this is currently some ability to query and update Flow collections in the API.
+This behaviour focuses on starting from a known Flow to which the collection is applied.
+There is no equivalent endpoints at the Source level.
+
+If chosen to also apply at the Flow level then need to consider is `flows_collected_by_id=a46c49f1-4764-42b9-9f91-f267a58903c4` too similar to `/flows/a46c49f1-4764-42b9-9f91-f267a58903c4/flow_collection`?
```
samdbmg: The distinction kinda comes up for the `role` discussion in #173 - the query string parameter lets you search for more things, for example `/flows?flows_collected_by_id=<uuid>&format=urn:x-nmos:format:video&codec=video/h264` which feels useful, while the `flows/<uuid>/flow_collection` is the canonical resource itself - you can edit it there

samdbmg (commented):
This looks good so far, and I think it makes sense to match across to Flows too. It's also useful for #173 because it gives you a way to find the collected Flows/Sources while querying by other things at the same time

samdbmg:
**TSC 2026-05-01**
Option 3 seems to be the preferred option here - @johnbilt will continue to write the associated spec change. Will tidy up the empty string behaviour in webhooks to match what is proposed here as well. Will also make it consistent across Sources and Flows.

samdbmg:
**TSC 2026-05-29**
Current state is this has the ADR but no spec change yet - next step is to write up the spec change

docs/adr/0049-source-collected_by-query-parameter.md
```diff
@@ -0,0 +1,103 @@
...
+* Option 4: Do nothing
+
+Flows:
+
+* Option A: Apply the same querying capabilities at both Source and Flow level
+* Option B: Only apply the new query capabilities at the Source level
+
+## Decision Outcome
+
+Recommendation: Option 3 since this already exists within the API as part of the webhooks, however this requires clarification of empty query parameters which is recommended to match the proposed behaviours.
```
j616: NIT: We've now accepted the chosen options
```suggestion
Chosen Options 3 and A since this already exists within the API as part of the webhooks, however this requires clarification of empty query parameters which is recommended to match the proposed behaviours.
```

docs/README.md
```diff
@@ -85,5 +85,6 @@ For more information on how we use ADRs, see [here](./adr/README.md).
 | [0044](./adr/0044-signalling-timeouts.md)                          | Signalling timeout periods                                                 |
 | [0046](./adr/0046-governance.md)                                   | Governance                                                                 |
 | [0048](./adr/0048-media-integrity.md)                              | Integrity model for media in TAMS, and when interacting with other systems |
+| [0049](./adr/0049-source-collected_by-query-parameter.md)                              | Ability to query for Sources without a collected_by value |
```
j616: Minor formatting nit: The `|` characters don't currently align with the other table rows. It's technically fine, but a touch messy in a code editor.

api/TimeAddressableMediaStore.yaml
```diff
@@ -412,6 +412,14 @@ paths:
           description: Filter on Source format.
           schema:
             $ref: 'schemas/content-format.json'
+        - name: collected_by_ids
```
j616, line 759: Do we want a `collects_ids` equivalent that goes in the opposite direction?

I think we also mused about a `collected_by_source/flow_ids` parameter that might query Flows/Sources by collections at the other level. But maybe thats a step too far...

j616 (commented):
Nice work. One question and a couple of minor suggestions.

docs/README.md
johnbilt: Done

api/TimeAddressableMediaStore.yaml
samdbmg: Is `GET /sources/<sourceId>` and then reading `source_collection` sufficient for that first part?

j616: I think its the other way around. I think you'd read the collected Sources (the one's you'd put in the hypothetical `collects_ids`) then read `collected_by`, then `GET /sources/<sourceId>` for each of those. So possible. But quite a few requests. One each for the IDs you're starting from. Then one each for the IDs identified in `collected_by`.

samdbmg: Ah, yes, I see that now - I suppose it might be useful to be able to query that way too

samdbmg: We discussed this in https://github.com/bbc/tams/issues/214#issuecomment-5217888262 and concluded that we've leave this for future work

samdbmg (approved):
LGTM, as discussed (I've taken the liberty of slipping in a minor fix to a typo I introduced in #225 )

samdbmg:
We discussed this on in https://github.com/bbc/tams/issues/214#issuecomment-5217888262 and the changes are non-functional history cleanup, so I'm going to go ahead and merge this

PR #185, opened 2026-04-16
Readme docs spelling mistakes

zhaolie-han:
# Details
Just some spelling mistakes in readme files.


j616:
Thank you for your contribution. Given the scope of these changes, we shall incorporate them into another PR rather than trouble you for an ICLA.

PR #186, opened 2026-04-17
Remove text about 404 if label not set

samdbmg:
# Details
404 would indicate the resource was not found, but in this case it was found, it just had no value.
Aligns with the other requests (e.g. `GET /flows/{flowID}/description`

I've put this as bugfix, although I can see at least [one implementation](https://github.com/sohonetlabs/rustytams/blob/main/tams-server/src/handlers/flow_props.rs#L74) that implements the behaviour as described, so I'd appreciate a second opinion on whether it's an API break

# Jira Issue (if relevant)
Jira URL: N/A

# Related PRs
N/A

# Submitter PR Checks
_(tick as appropriate)_

# Reviewer PR Checks
_(tick as appropriate)_

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -747,7 +747,7 @@ paths:
         "200":
           $ref: '#/components/responses/trait_resource_info_head_200'
         "404":
-          description: The requested Source does not exist, or does not have a label set.
+          description: The requested Source does not exist.
     get:
       summary: Source Label
       description: Returns the Source label property. This should be a very short, human-readable label that may be displayed in listings of Sources.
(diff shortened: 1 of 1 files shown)


j616 (approved):
> I've put this as bugfix, although I can see at least [one implementation](https://github.com/sohonetlabs/rustytams/blob/main/tams-server/src/handlers/flow_props.rs#L74) that implements the behaviour as described, so I'd appreciate a second opinion on whether it's an API break

We generally consider version bumps from the perspective of the client. With that in mind, I think I'd consider it a feature bump at most. But I think it's a bit moot given the next release will be a minor release anyway

PR #222, opened 2026-07-02
Support specification of presigned/non-presigned URLs on put_urls

### LINKED ISSUE #221: Unsigned S3 URI not available in response for /storage endpoint
**Context**
In TAMS managed storage mode, when a request is made to get storage endpoints, there is currently no option to request raw S3 URIs, and the response defaults to pre-signed URLs. For systems that require low latency and perform bulk storage requests, signing each endpoint on the fly by the TAMS server might add extra delay. In the `/segments` GET path, requesting raw S3 URIs is supported, so for symmetry it makes sense to provide clients an option to request raw S3 URIs from `/storage` as well. Systems that want to restrict themselves to IAM roles due to security compliance would also want to get S3 URIs and not pre-signed URLs.

**Description**
In the response of `/storage`, `put_url` will become `put_urls`, and based on a query param, the client can request raw S3 URIs instead of pre-signed URLs.

**Use Case**
1. Media processing jobs that use GStreamer pipelines where having as little latency as possible is best
2. Systems that want to use IAM roles rather than pre-signed URLs for security reasons

j616:
# Details
* Adds the ability to explicitly select the use of pre-signed `put_url`s.
* Adds the ability to explicitly signal where `put_url`s are pre-signed or not
* Additional clarity on the authentication methods used by `put_url`s and `get_urls`

# Issue (if relevant)
GitHub Issue: #221 
Jira ticket: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5588

# Related PRs

# Submitter PR Checks

- [x] PR completes task/fixes bug
- [x] API version has been incremented if necessary
- [x] PR added to Jira Issue (if relevant)

# Reviewer PR Checks

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/schemas/flow-storage-post.json
+++ b/api/schemas/flow-storage-post.json
@@ -21,6 +21,10 @@
         "content_type": {
             "description": "The `content_type` to use for the Objects. This parameter MUST only be set where requesting storage for initialisation segments in media formats which require them, and where the mime-type of those initialisation segments differs to that of the media segments. Assumed to be the `container` type of the Flow if not set.",
             "$ref": "mime-type.json"
+        },
+        "presigned": {
+            "description": "If set to `true`, the `put_url`'s in the response will be presigned. If set to `false`, the `put_url`'s in the response will not be presigned. If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to generate a large number of pre-signed URLs. Services may choose their own default. If Clients only support one mode of operation, they SHOULD specify this parameter.",
+            "type": "boolean"
         }
     },
     "not": {
--- a/api/schemas/flow-storage-post.json
+++ b/api/schemas/flow-storage-post.json
@@ -23,7 +23,7 @@
             "$ref": "mime-type.json"
         },
         "presigned": {
-            "description": "If set to `true`, the `put_url`'s in the response will be presigned. If set to `false`, the `put_url`'s in the response will not be presigned. If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to generate a large number of pre-signed URLs. Services may choose their own default. If Clients only support one mode of operation, they SHOULD specify this parameter.",
+            "description": "If set to `true`, the `put_url`s in the response will be presigned. If set to `false`, the `put_url`s in the response will not be presigned. If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to generate a large number of pre-signed URLs. Services may choose their own default. If Clients only support one mode of operation, they SHOULD specify this parameter.",
             "type": "boolean"
         }
     },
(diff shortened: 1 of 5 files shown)


api/schemas/flow-storage-post.json
```diff
@@ -21,6 +21,10 @@
         "content_type": {
             "description": "The `content_type` to use for the Objects. This parameter MUST only be set where requesting storage for initialisation segments in media formats which require them, and where the mime-type of those initialisation segments differs to that of the media segments. Assumed to be the `container` type of the Flow if not set.",
             "$ref": "mime-type.json"
+        },
+        "presigned": {
+            "description": "If set to `true`, the `put_url`'s in the response will be presigned. If set to `false`, the `put_url`'s in the response will not be presigned. If `presigned` is set to `false`, the response from the service could be substantially faster if it is not required to generate a large number of pre-signed URLs. Services may choose their own default. If Clients only support one mode of operation, they SHOULD specify this parameter.",
```
jamie-bbc: minor nit: `put_url`s shouldn't have an apostrophe as it is not possessive or a contraction.

jamie-bbc (approved):
I'm happy with this from what I've read - LGTM! Approving this, but happy if you want a second opinion.
Just a nit that each time `put_url`s is mentioned there shouldn't be an apostrophe.

api/schemas/flow-storage-post.json
j616: Thank you. Fixed.

Akshay-Narayan-Pai (commented):
Is put_url becoming put_urls ? Then are we having both in API then slowly deprecating ? Or here as per the changes it is seeming like get_urls is becoming get_url but I don't see a openapi spec key change either

j616:
> Is put_url becoming put_urls ? Then are we having both in API then slowly deprecating ? Or here as per the changes it is seeming like get_urls is becoming get_url but I don't see a openapi spec key change either

The approach I've taken in this PR is to keep the structure of both `get_urls` and `put_url` the same. I've added a `presigned` flag to the `/storage` POST method request body to allow clients to select pre-signed URLs, and to the response body to signal if a URL is presigned. 

My intent is to keep this change backwards compatible as this is a core part of the API. I also feel like a list of URLs makes more sense on the read side than the write side. On the write side, the client either specifically knows which Storage Backend they wish to use, or they are happy to accept the default. If they have a preference for pre-signed/not then they will signal that by setting the `presigned` parameter. On the read side, a consuming Client often can't know up front which Storage Backend(s) Objects are available in. And in some cases, such as zero-copy-edit, may vary throughout the Flow. We do allow consuming Clients to filter the URLs returned where they have that knowledge, but we have to provide the ability to return a list so Clients can select based on what is available and information such as logical/geographical location. One could argue that its superfluous to return both presigned and non-presigned variants for the same Object Instance on the read side where a preference hasn't been stated, and perhaps we should call that out in the spec. But I think that's an implementation decision to some extent.

Akshay-Narayan-Pai:
makes sense thanks for explanation, it was somewhat confusing with apostrophe

samdbmg:
**TSC 2026-07-24**
Merge.

PR #223, opened 2026-07-03
Explicitly specify default ordering of segments listing

### LINKED ISSUE #219: [BUG] Segments are not explicitly described as returned in time order
**Describe the bug**
The [documentation](https://bbc.github.io/tams/main/index.html#/operations/GET_flows-flowId-segments) for `GET /flows/<flowid>/segments` doesn't explicitly say that segments should be returned in time order, but implies it by the existence of `?reverse_order`

**Expected behavior**
While most practical implementations return the segments in time order, many players/readers rely on being able to assume the list is sorted, so it should be called out explicitly.

**Suggested fix**
Call out explicitly that the results should be sorted in order

**Additional context**
Found as part of https://github.com/google/virtual-broadcast-production-assistant/issues/314

j616:
# Details
Explicitly specify default ordering of segments listing

# Issue (if relevant)
GitHub Issue: #219 

# Related PRs

# Submitter PR Checks

# Reviewer PR Checks

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1863,7 +1863,7 @@ paths:
 
         It may also use a subset of the samples in the Media Object, and if the `include_object_timerange=true` parameter is set, the object's timerange will also be returned to aid identifying which samples to skip.
 
-        Segments are sorted by ascending timerange by default.
+        Segments are sorted by timerange, in ascending order by default.
 
         Clients should use the pagination options to limit the results to a timerange and/or count.
         Service implementations may also limit the results returned.
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1863,6 +1863,8 @@ paths:
 
         It may also use a subset of the samples in the Media Object, and if the `include_object_timerange=true` parameter is set, the object's timerange will also be returned to aid identifying which samples to skip.
 
+        Segments are sorted by ascending timerange by default.
+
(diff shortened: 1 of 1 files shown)


api/TimeAddressableMediaStore.yaml
```diff
@@ -1863,6 +1863,8 @@ paths:
 
         It may also use a subset of the samples in the Media Object, and if the `include_object_timerange=true` parameter is set, the object's timerange will also be returned to aid identifying which samples to skip.
 
+        Segments are sorted by ascending timerange by default.
```
samdbmg: Nit: Does this read better?
```suggestion
        Segments are sorted by timerange, in ascending order by default.
```

Also do we need a reminder that it doesn't matter whether you sort by timerange start or end: the result is the same because they cannot overlap?

samdbmg (approved):
LGTM

api/TimeAddressableMediaStore.yaml
j616: I was thinking that. I couldn't decide if it was stating the obvious or not.

j616: Thinking about this over the weekend, I've added in a note on the sort key. Further explanation isn't a bad thing.

j616: I have also included a note about edge cases involving clusivity markers.

samdbmg (approved):
LGTM - I agree on your point about more explanation - and good thought on edge cases

PR #224, opened 2026-07-07
Specify the sorting of resource listings

### LINKED ISSUE #189: Add the ability to sort Flow/Source listings
**Description**
It would be useful to be able to sort the list of Flows and Sources to aid in building UIs - e.g. by providing a query string parameter

**Use Case**
As an NLE TAMS plugin vendor, I'd like to present a useful list of Flows and Sources to the user. However returning them in an unstable order or sorting by pulling the entire list and then sorting clientside lead to inefficiency or bad user experience

**Action**
- [ ] Write an ADR covering these broad options:

1. No sorting - all clientside
2. Sorting by any of the query string parameters already in the spec
3. Fixed default sort by most recent `metadata_updated`
4. Optional sort by `metadata_updated`

- [ ] Draft the spec changes for one of the simple options

**Outcome**
- ADR PR
- Possibly with spec change...

j616:
# Details
TAMS hasn't previously specified how resource listings should be sorted. This PR includes an ADR on options for how resource listings should be sorted.

# Issue (if relevant)
GitHub Issue: #189 
Jira: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5486

# Related PRs

# Submitter PR Checks

# Reviewer PR Checks

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -135,8 +135,31 @@ paths:
           schema:
             default: false
             type: boolean
+        - $ref: '#/components/parameters/trait_resource_paged_key'
+        - $ref: '#/components/parameters/trait_paged_limit'
       responses:
         "200":
+          headers:
+            Link:
+              description: Provides references to cursors for paging. Only the 'rel' attribute with value 'next' is currently supported. If 'next' is not present then it is the last page.
+              schema:
+                type: string
+            X-Paging-Limit:
+              description: Identifies the current limit being used for paging. This may not match the requested value if the requested value was too high for the service implementation
+              schema:
+                type: integer
+            X-Paging-Count:
+              description: The number of items in the returned data set.
+              schema:
+                type: integer
+            X-Paging-Reverse-Order:
+              description: The items are returned in reverse order.
+              schema:
+                type: boolean
+            X-Paging-NextKey:
+              description: Opaque string that can be supplied to the `page` query parameter to get the next page of results.
+              schema:
+                type: string
           $ref: '#/components/responses/trait_resource_listing_head_200'
     get:
       summary: Storage Backend Information
@@ -155,9 +178,32 @@ paths:
           schema:
             default: false
             type: boolean
+        - $ref: '#/components/parameters/trait_resource_paged_key'
+        - $ref: '#/components/parameters/trait_paged_limit'
       responses:
         "200":
           description: ""
+          headers:
+            Link:
+              description: Provides references to cursors for paging. Only the 'rel' attribute with value 'next' is currently supported. If 'next' is not present then it is the last page.
+              schema:
+                type: string
+            X-Paging-Limit:
+              description: Identifies the current limit being used for paging. This may not match the requested value if the requested value was too high for the service implementation
+              schema:
+                type: integer
+            X-Paging-Count:
+              description: The number of items in the returned data set.
+              schema:
+                type: integer
+            X-Paging-Reverse-Order:
+              description: The items are returned in reverse order.
+              schema:
+                type: boolean
+            X-Paging-NextKey:
+              description: Opaque string that can be supplied to the `page` query parameter to get the next page of results.
+              schema:
+                type: string
           content:
             application/json:
               schema:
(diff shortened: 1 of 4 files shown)


docs/adr/0052-sort-listings.md
```diff
@@ -0,0 +1,110 @@
+---
+status: "proposed"
+---
+# Options for specifying the sorting of Listings
```
samdbmg: Nit: We've ended up with an unwritten convention that the file names and titles broadly match: is it worth lining these up?

docs/adr/0052-sort-listings.md
```diff
@@ -0,0 +1,110 @@
...
+* Option 2 - Define a single key all implementations should sort listings by
+* Option 3 - Define a set list of sorting keys clients may select from
+* Option 4 - Allow clients to select any resource parameters to sort by
+
+## Decision Outcome
+
+Chosen option: "Option 3 - Define a set list of sorting keys clients may select from".
+Although some endpoints only have a single obvious option for sorting.
+These will use "Option 2 - Define a single key all implementations should sort listings by".
+This provides the most clarity and flexibility to implementations without placing undue burdon to Service implementations.
```
samdbmg: Nit: Typo
```suggestion
This provides the most clarity and flexibility to implementations without placing undue burden to Service implementations.
```

docs/adr/0052-specifying-the-sorting-of-listings.md
```diff
@@ -0,0 +1,110 @@
...
+## Decision Outcome
+
+Chosen option: "Option 3 - Define a set list of sorting keys clients may select from".
+Although some endpoints only have a single obvious option for sorting.
+These will use "Option 2 - Define a single key all implementations should sort listings by".
+This provides the most clarity and flexibility to implementations without placing undue burdon to Service implementations.
+All endpoints will have a `reverse_order` query parameter, matching that which currently exists on the `/flows/{flowId}/segments` endpoint.
+
+Datetime keys shall be sorted newest-first by default.
+String keys shall be sorted ascending alphabetically by default.
```
samdbmg, line 40: Do we need to specify (or preferably, cite) a lexicographical ordering to avoid uncertainty for casing and numbers?

samdbmg (approved):
Looks good overall - couple of thoughts inlined

docs/adr/0052-specifying-the-sorting-of-listings.md
j616: I left this out because I think lexographical can often result in lower case being grouped and upper case being grouped, if I'm not mistaken. At least, how they are defined in programming languages differs. With the likes of Python requiring you to go as far a key such as `sorted("This is a test string".split(), key=str.casefold)` which may be just as un-clear in terms of how numbers/none alpha-numeric chars are handled. I felt stating "alphabetically" better conveyed sorting such that `a` is next to `A` without having to dig into the peculiarities of languages.

api/TimeAddressableMediaStore.yaml
danjhd, line 1: Do we need to add the `X-Paging-Reverse-Order` header too to be consistent with the Segments endpoint?

j616: Yes! Fixed in d8abaf8.

api/TimeAddressableMediaStore.yaml
```diff
@@ -835,6 +985,26 @@ paths:
...
+          in: query
+          description: Return Flows in reverse order.
+          schema:
+            default: false
+            type: boolean
+        - name: sort_by
+          in: query
+          description: |
+            Parameter to sort Flows by.
+            `created` and `metadata_updated` sort most recent first by default.
```
danjhd, line 1195: I know we are using metadata_updated here to be consistent with the field name, but i believe the only reason the field name here is `metadata_updated` instead of just `updated` (as it is on Sources) is historic. So i wonder whether we considered using just `updated` here to be consistant in the enums for the query parameter at least? Just a thought.

j616: We have `metadata_updated` here to distinguish it from `segments_updated`. I think I prefer (to use Sam's terminology, strong preference weekly held) local consistency to global consistency in this case such that the param values are very explicit about which sort key they are selecting.

danjhd: Fair enough, cheers.

danjhd: Happy to resolve this (i don't have the permissions)

samdbmg:
**TSC 2026-07-24**
Merge.

PR #225, opened 2026-07-10
Make `role` optional in collections, capture the same information elsewhere

### LINKED ISSUE #188: Role name description should be more specific
**Description**
The description and recommendations around the values of "Role" in collections are currently quite vague. This may impeed interoperability. We should expand our recommendations/examples in this space, potentially referencing existing specifications in this space.

**Use Case**
As an ingest vendor, I would like to understand what should appear in the `role` field when I create multi-essence Flows.

As a store browser/Flow player, I would like to understand how to display to my user the contents of the multi-essence Flow/Source I'm looking at.

samdbmg:
# Details
- Adds an ADR exploring what `role` should be used for in collections, and concluding it should become optional
- Makes the corresponding change to the API
- ...and examples
- Adds an ADR considering where to capture the "editorial purpose" of a Source or Flow (something `role` has been used for), and corresponding appnote, tags and examples

# Issue (if relevant)
GitHub Issue: Fixes #188 
Also tracked as https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5509

# Related PRs
Replaces #173 since that PR had got rather messy!

# Submitter PR Checks

# Reviewer PR Checks

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs

### DIFF
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -923,6 +923,10 @@ paths:
                   summary: Multi-essence Source
                   description: Multi-essence Sources collect multiple Sources of different formats under one Source ID.
                   externalValue: examples/source-get-200-multi.json
+                tags:
+                  summary: Source with tags
+                  description: Source example with tags set matching associated Flow
+                  externalValue: examples/source-get-200-tags.json
         "404":
           description: The requested Source does not exist.
   /sources/{sourceId}/tags:
@@ -1542,6 +1546,9 @@ paths:
                 video_vfr:
                   summary: Video Flow - H.264 Codec, Variable Frame Rate
                   externalValue: examples/flow-get-200-video-h264-vfr.json
+                video_cleanfeed:
+                  summary: Video Flow - H.264 Codec, with Editorial Purpose Tag
+                  externalValue: examples/flow-get-200-video-h264-cleanfeed.json
                 audio:
                   summary: Audio Flow - AAC Codec
                   externalValue: examples/flow-get-200-audio-aac.json
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1593,7 +1593,7 @@ paths:
         Service implementations SHOULD verify that Flow metadata is compatible with the associated Source.
         Service implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict.
         Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
-        Examples of conflicting metadata include `format` not matching, or the `priority` or `role` in `source_collection` and `flow_collection` not matching.
+        Examples of conflicting metadata include `format` not matching, or `role` or ordering in `source_collection` and `flow_collection` not matching.
         It may also be possible for service implementations to detect some instances where multiple Flows should not be considered of the same Source, such as audio Flows with different numbers of tracks.
         Further guidance on when Flows/Sources may be considered the same/different may be found in the [Practical Guidance for Media](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.0._Practical_Guidance_for_Media.html) section of AMWA MS-04.
 
@@ -2021,7 +2021,7 @@ paths:
           $ref: '#/components/responses/trait_resource_info_head_404'
     get:
       summary: Flow Collection
-      description: Returns the Flow collection property. A list of Flows that are collected together by this Flow, sorted in ascending `priority` order (where no `priority` sorts last).
+      description: Returns the Flow collection property. An ordered list of Flows that are collected together by this Flow.
       operationId: GET_flows-flowId-flow-collection
       tags:
         - Flows
@@ -2039,7 +2039,7 @@ paths:
     put:
       summary: Create or Update Flow Collection
       description: |
-        Create or update the Flow collection property. A list of Flows that are collected together by this Flow.
+        Create or update the Flow collection property. An ordered list of Flows that are collected together by this Flow.
 
         Service implementations SHOULD verify that Flow metadata is compatible with the associated Source.
         Service implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict.
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1593,7 +1593,7 @@ paths:
         Service implementations SHOULD verify that Flow metadata is compatible with the associated Source.
         Service implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict.
         Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
-        Examples of conflicting metadata include `format` not matching, or the `priority` in `source_collection` and `flow_collection` not matching.
+        Examples of conflicting metadata include `format` not matching, or the `priority` or `role` in `source_collection` and `flow_collection` not matching.
         It may also be possible for service implementations to detect some instances where multiple Flows should not be considered of the same Source, such as audio Flows with different numbers of tracks.
         Further guidance on when Flows/Sources may be considered the same/different may be found in the [Practical Guidance for Media](https://specs.amwa.tv/ms-04/releases/v1.0.0/docs/3.0._Practical_Guidance_for_Media.html) section of AMWA MS-04.
 
--- a/api/TimeAddressableMediaStore.yaml
+++ b/api/TimeAddressableMediaStore.yaml
@@ -1586,7 +1586,7 @@ paths:
         Service implementations SHOULD verify that Flow metadata is compatible with the associated Source.
         Service implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict.
         Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
-        Examples of conflicting metadata include `format` not matching, or the `role` in `source_collection` and `flow_collection` not matching.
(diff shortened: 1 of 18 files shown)


api/schemas/collection-item.json
```diff
@@ -3,16 +3,20 @@
...
             "$ref": "uuid.json"
         },
+        "priority": {
+            "description": "A number indicating the position of this element in the collection, provided to allow stable sorting in UIs and the presentation of defaults. Implementations SHOULD return results sorted by this field in ascending order, and items with no `priority` set sorted last. Where multiple elements have the same priority, their order is implementation-defined.",
+            "type": "integer",
+            "minimum": 0
+        },
         "role": {
-            "description": "A human-readable role of the element in this collection (e.g. 'R' to denote a right audio channel in a collection of mono audio Sources)",
+            "description": "An optional string describing the purpose of this element in the collection, primarily intended to be human-readable.",
```
j616:
```suggestion
            "description": "The purpose of this element in the collection, primarily intended to be human-readable.",
```
NIT: `optional` and `string` feel redundant as they're in the machine readable part of the spec

api/examples/flow-get-200-multi.json
```diff
@@ -10,11 +12,16 @@
     "flow_collection": [
         {
             "id": "4f79cfd1-c057-47f4-8e4d-1b126ca7bf34",
-            "role": "video"
+            "priority": 1
```
j616: The example priorities seem to be unique to Flow type. Should they be unique among the collection? One use case is track ordering in an NLE, and I don't think they are always grouped by type? I also wonder if there are other interfaces that are more generic that might group everything together.

api/TimeAddressableMediaStore.yaml
```diff
@@ -1116,7 +1123,7 @@ paths:
         Service implementations SHOULD verify that Flow metadata is compatible with the associated Source.
         Service implementations MAY accept modification/addition of parameters, and reflect such changes in the Source, where it will not bring any Flows of the Source into conflict.
         Where metadata would result in any Flow of the Source coming into conflict, the request SHOULD be rejected with a 400 response.
-        Examples of conflicting metadata include `format` not matching, or the `role` in `source_collection` and `flow_collection` not matching.
+        Examples of conflicting metadata include `format` not matching, or the `priority` in `source_collection` and `flow_collection` not matching.
```
j616: I think the conflicting "role" is still appropriate here? It's optional but still shouldn't conflict.

docs/adr/0047-roles-in-collections.md
```diff
@@ -0,0 +1,104 @@
...
+* Items in a collection serve a variety of different purposes, and it is useful to have some commonality.
+* Common selection and display conventions are useful as content is read from TAMS across multiple clients and services.
+
+## Considered Options
+
+* Option 1: Make `role` a semi-structured field containing editorial purpose, media type and other details
+* Option 2: Replace `role` with additional controlled fields
+* Option 3: Use `role` as editorial purpose, use other queries for Flow/Source properties
+* Option 4a: Capture editorial purpose elsewhere, use `role` as an optional label
+* Option 4b: Capture editorial purpose elsehwere, use `role` as an optional label, add a `priority` for sorting
```
j616:
```suggestion
* Option 4b: Capture editorial purpose elsewhere, use `role` as an optional label, add a `priority` for sorting
```

docs/adr/0053-editorial-purpose-tag.md
```diff
@@ -0,0 +1,88 @@
...
+* Bad, because the component types are represented as hex bytes (for carriage in the SDT table) and would need human-readable names
+* Bad, because many of the types also include technical characteristics of content, making them unsuitable for use in Sources
+* Bad, because the list is focused on distribution, so cannot contain aspects such as clean-feed video
+
+### Option 1b: Represent editorial purpose as a field using MPEG-DASH roles
+
+As Option 1a, using MPEG-DASH role values instead.
+
+MPEG-DASH contains a role attribute for an `AdaptationSet`, which describes the purose of that particular track.
+A number of values for that attribute are given in the specification (see ISO/IEC 23009-1:2022 section 5.8.5.5), covering the `main` content along with others such as `alternate`, `supplementary`, `commentary`, `description`, etc.
```
j616: NIT: Link?

docs/appnotes/0020-editorial-purpose.md
```diff
@@ -0,0 +1,33 @@
+# 0020: Editorial Purpose
+
+## Abstract
+
+An `editorial_purpose` tag is defined in [AppNote 0003: Tag Names](./0003-tag-names.md) to give a hint about what a particular Source or Flow contains and how it could be used.
+The tag's values are primarily intended to be human-readable and uncontrolled, however this list provides suggested values and their meanings to allow some commonality.
+Much like other tag definitions in TAMS, the list can be amended and will likely grow over time.
```
j616: Nit: Do we want to provide any guidance on implemnentation specific values, akin to that in the tags listing? e.g. value starts with `_<implementationName>....`

docs/appnotes/0003-tag-names.md
```diff
@@ -86,6 +86,13 @@ Status: **Deprecated**
 Replaced by the `created` field in Flow metadata.
 Contains the ISO formatted date-time when a Flow was created.
 
+### editorial_purpose
+
+Status: **Proposed**
+
+Captures the editorial purpose of a piece of content, giving a hint about what it's for and what can be found inside.
```
j616, line 93: Do we need to call out if these are single values, or if lists are allowed?

docs/appnotes/0020-editorial-purpose.md
```diff
@@ -0,0 +1,33 @@
...
+
+For audio SMPTE ST377-41 provides a rich vocabulary for audio content, which can be used directly by TAMS.
+A number of common examples are reproduced here: for the full list see _Section 5.4 MCA Content_ in [ST377-41:2023](https://pub.smpte.org/doc/st377-41/): by convention TAMS uses the "MCA Content" name (rather than the Symbol value), lowercased.
+
+_Note that in general, audio Flows and Sources should also use a [language code](https://github.com/bbc/tams/blob/main/docs/appnotes/0003-tag-names.md#language_code) label to allow clients to identify the language used._
+
+| Name | Description |
+| ---- | ----------- |
+| `primary` | Primary, or default audio for a piece of content, e.g. that will be edited or distributed on to the audience. |
+| `descriptive video` | Audio description track, mixed with primary programme content for presentation to visually impaired audiences. |
```
j616: NIT: The video values above use underscores, these have spaces. I suspect underscores should be preferred to reduce the potential need to escape values

docs/appnotes/0025-editorial-purpose.md
```diff
@@ -0,0 +1,33 @@
+# 0020: Editorial Purpose
+
+## Abstract
+
+An `editorial_purpose` tag is defined in [AppNote 0003: Tag Names](./0003-tag-names.md) to give a hint about what a particular Source or Flow contains and how it could be used.
+The tag's values are primarily intended to be human-readable and uncontrolled, however this list provides suggested values and their meanings to allow some commonality.
+Much like other tag definitions in TAMS, the list can be amended and will likely grow over time.
+
+## Editorial Purposes
```
j616, line 9: Should we have a data section here with values for subtitles, in particular? I'm thinking `sdh` for Subtitles for the Deaf and Hard of Hearing, and `transcript` for subtitles without any descriptive info regarding sound effects etc.

api/examples/flow-get-200-multi.json
samdbmg: One of the "rules" is that multiple elements can have the same priority (and that ordering is implementation-defined) - otherwise priority unset becomes a weird special case. I suppose I can see arguments both ways in practice: setting equivalent priority among different Flow types on the basis it naturally groups some together, and suggesting they all be different so the ordering is entirely stable. I suppose you could achieve the former with something like:
- Flow 1: Video - Priority 1
- Flow 2: Audio - Priority 2
- Flow 3: AD - Priority 10

What do you think - is the grouping of like Flows useful, or something we should do "properly" in another way?

docs/appnotes/0003-tag-names.md
samdbmg: Lists is an exciting possibility I hadn't previously considered, but maybe we should see if that shakes out in practical usage?

docs/appnotes/0020-editorial-purpose.md
samdbmg: I was thinking about that - I'm trying to directly copy ST377-41 but we could expand the lowercasing rule above to include spaces to underscores

api/examples/flow-get-200-multi.json
samdbmg: After some discussion: there's no particular reason we need `priority` to preserve ordering: we could just say that the collection is ordered. You can only edit the list by submitting the whole thing in a `PUT` anyway so this would probably make sense. Internally implementations may need to capture an ordering (depending on the underlying storage) but that's an implementation detail.

docs/appnotes/0020-editorial-purpose.md
```diff
@@ -0,0 +1,40 @@
...
+| `live_commentary` | Commentary track, where sent separately (e.g. from a sports event).  |
+| `music_and_effects` | Music and Effects track, where sent separately. |
+| `audio` | In a simple A/V mux that only contains the audio and video for an asset, the `role` provides little additional information, and calling it "audio" may be sufficient. |
+
+## Data
+
+| Name | Description |
+| ---- | ----------- |
+| `sdh` | Subtitles for the Deaf and Hard of Hearing, including descriptive information of audio cues and other non-dialogue sound |
+| `transcript` | Subtitles containing only the spoken words |
```
j616:
```suggestion
| `transcript` | Transcript containing only the spoken words |
```
NIT: Aware this is a bit self-referential. But I think a transcript often wouldn't be considered subtitles as it doesn't contain any of the display metadata one would expect alongside the text of subtitles (aka closed captions)

j616 (approved):
One Nit. Otherwise LGTM

PR #227, opened 2026-07-10
Tag status review

### LINKED ISSUE #206: Review tags for elevation to the core spec
**Context**
Tags are often used in TAMS to explore features before committing to baking them into the spec. Some tags in this class have been in use for some time now (e.g. `flow_status`). We should review our tags listing and elevate tags to the core spec where appropriate.

**Actions**
* Review tags for potential elevation to the core spec
* Write and ADR documenting the decision process for candidates for elevation
* Review this with the TSC, and wider community as appropriate
* Make the relevant spec changes, and changes to AppNote0003

**Outcome**
Tags have been reviewed for elevation to the core spec, and elevated as appropriate.

j616:
# Details
This PR is the result of a review of the status of tags in the AppNote0003 registry. The following changes have been made:
* Elevate the `flow_status` tag to the core spec, as the `status` Flow parameter
  * This tag is widely used in implementations and seen as stable
* Elevate the `auth_classes` and `language_code` tags from `Experimental` to `Proposed`
  * These tags have existed for some time without changes and have had their status adjusted to reflect this
* Deprecate the `originating_id` and `originating_timerange` tags
  * These tags have largely been replaced by the `first_referenced_by_flow` parameter in Object metadata
* Marked `hls_exclude` and `hls_segments` tags as Implementation specific
  * These parameters have been used in demos and have seen limited use in implementations, but are not widely used. We also strongly recommend the use of TAMS directly and not via HLS adaptation layers as these do not support TAMS timing or partial Object use. Results may be unexpected where HLS is used.

# Issue (if relevant)
GitHub Issue: #206 
Jira ticker: https://jira.dev.bbc.co.uk/browse/CLOUDFIT-5487

# Related PRs

# Submitter PR Checks

# Reviewer PR Checks

- [ ] Code is easy to follow
- [ ] PR size is sensible

# Info on PRs


johnbilt:
Elevating the flow status tag makes a lot of sense and is being used now in a number of implementations where people need to understand whether a flow is ingesting.  There have been some discussions around whether there is more functionality to add around flow status rather than just elevating the existing tag to the core specification.   These include:

1.  The role of the store in managing the flow_status field.  Currently this is down to the ingest tool to manage the state of the field, and while this is the best practise we do see ingests which have finished but the tag has not been cleared.  One option is for the store take responsibility for automatically closing flows which have had no new segments created after a period of time (eg 15 minutes) where the status has not been set to 'awaiting_content'.  Similarly, should the store take any responsibility for setting the status in the first place, eg 'awaiting_content' could be set automatically on the flow creation and then 'ingesting' set when the first segments are received.  Combined with the auto closing could mean this works regardless of the capabilities of the ingest solution but does put additional work on the store to manage

2.  There has been a request to roll up the flow status into the source level and specifically to the top level source.  In a world of source first discovery then currently the only way to know if a piece of content is ingesting is to walk the whole tree and evaluate each flow to find out if there are any still ingesting.  Having the status ripple up to the top of the data structure would be useful, but again puts more things for the store to manage, plus needs rules defining about how to handle different statuses on different flows.

j616:
> 1. The role of the store in managing the flow_status field.  Currently this is down to the ingest tool to manage the state of the field, and while this is the best practise we do see ingests which have finished but the tag has not been cleared.  One option is for the store take responsibility for automatically closing flows which have had no new segments created after a period of time (eg 15 minutes) where the status has not been set to 'awaiting_content'.  Similarly, should the store take any responsibility for setting the status in the first place, eg 'awaiting_content' could be set automatically on the flow creation and then 'ingesting' set when the first segments are received.  Combined with the auto closing could mean this works regardless of the capabilities of the ingest solution but does put additional work on the store to manage

I feel like this might cross the line we've traditionally held that TAMS is about advertising available content, as far as possible. We've traditionally steered away from active behaviour like this. And I'm not sure this is in the same ballpark as garbage collection of Objects, for example. But I wonder if it is similar in scope to [AppNote0019 retention management](https://github.com/bbc/tams/blob/main/docs/appnotes/0019-implementing-retention-management.md). I wonder if an App Note on how this parameter may be sanitized via (potentially external) automation processes is the better approach, which fits more with existing patterns.

> 2. There has been a request to roll up the flow status into the source level and specifically to the top level source.  In a world of source first discovery then currently the only way to know if a piece of content is ingesting is to walk the whole tree and evaluate each flow to find out if there are any still ingesting.  Having the status ripple up to the top of the data structure would be useful, but again puts more things for the store to manage, plus needs rules defining about how to handle different statuses on different flows.

I'd be cautious about this behaviour. It feels like while this adds some convenience, it leads to more opportunity for inconsistencies. Its quite possible for only some, not all, Flows of a given Source to be currently ingesting. So marking the top level Flow/Source as ingesting may not be appropriate. I wonder if this would better solved by adding a query param to filter by `status`, and then using that in combination by the upcoming `collected_by` filter?

j616:
I've added a query parameter to filter Flows on `status` in 421a4c3

api/schemas/flow-status.json
```diff
@@ -0,0 +1,11 @@
+{
+    "title": "Flow Status",
+    "description": "The current ingest status of a Flow. Available values are as follows. `awaiting_content` - Flow is expecting, but not currently receiving content. `ingesting` - Content is currently being ingested. `replication_in_progress` - Content is currently being ingested to this Flow from another Service Instance via a replication process. `closed_complete` - Flow is complete and will not receive any more content. NOTE: Because this parameter is maintained/updated by a Client, the value of this parameter is only indicative and not authoritative. If the Client ingesting the content becomes unavailable, it may leave this parameter in an incorrect state. Clients should aim to tidy up this state appropriately once they recover.",
```
samdbmg: Nit: Does this render better if you add newlines after each item?

api/schemas/flow-core.json
```diff
@@ -81,6 +81,10 @@
             "type": "integer",
             "minimum": 0
         },
+        "status": {
+            "description": "The current ingest status of a Flow. Available values are as follows. `awaiting_content` - Flow is expecting, but not currently receiving content. `ingesting` - Content is currently being ingested. `replication_in_progress` - Content is currently being ingested to this Flow from another Service Instance via a replication process. `closed_complete` - Flow is complete and will not receive any more content. NOTE: Because this parameter is maintained/updated by a Client, the value of this parameter is only indicative and not authoritative. If the Client ingesting the content becomes unavailable, it may leave this parameter in an incorrect state. Clients should aim to tidy up this state appropriately once they recover.",
```
samdbmg: Doesn't this inherit the description of the `$ref` and make this one redundant?

samdbmg (approved):
Looks good, overall.

Regarding the question about automatic updates, I suppose we could put some language in somewhere about how that isn't part of the core spec, but that
> Implementers MAY consider adding more active management of the status of Flows where it isn't more actively controlled by clients, as part of more general lifecycle management capability, in line with the approach described for [retention management](https://github.com/bbc/tams/blob/main/docs/appnotes/0019-implementing-retention-management.md)

...or something snappier

api/schemas/flow-status.json
j616: Just checked. New lines have no affect on the rendered docs so I've left them out.

